PR #4, opened 2014-01-22
robotMotorGui default configuration file

elen4:
Since the ResourceFinder now searches "robot" directories before "context" ones, it is possible to provide a default configuration file for the robotMotorGui within a context. Current solution, with hard-coded default values, was incomplete, as actually a large number of values are needed for the gui to be fully operational (i.e., for "home" buttons to work), so I reverted related all commits. I changed the default robot name to be icubSim, so that it's easier to understand whether the robot-specific or default file has been loaded (and also because I assume that if a robot is on installed on a machine, the user is working with the simulator; anyway, I think that the simulator at some point should have a set of configuration files that match the real robots ones and that are found in the same way).

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -46,4 +46,5 @@ add_subdirectory(iCubStartup)
 add_subdirectory(dataSetPlayer)
 add_subdirectory(simFaceExpressions)
 add_subdirectory(fingersTuner)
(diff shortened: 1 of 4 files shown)


pattacini:
`robotMotorGui` is going to disappear from within `icub-main` and enter the world of  `yarp` under the livery of `yarpmotorgui`.

closing...

Issue #12, opened 2014-03-27
Add diagnostic capabilities to icub skin


JoErNanO:
The skin should be capable to relay status codes and error messages. @maggia80 and I are working on this.

JoErNanO:
Closed as diagnostics were merged by @barbalberto in cbc0fa0.

PR #17, opened 2014-04-28
Skin diagnostics #12

JoErNanO:
Skin diagnostics works as explained here: https://github.com/robotology/icub-main/issues/12#issuecomment-41506010


barbalberto:
Change the parameters as @randaz81 suggested (removed the patchesIdList)

Still to do:

-  use the "all is well" messages as heartbeat check
- decimate error messages in case of persistant errors

Tested on the blue iCub and it is fine with both eth (arms) and can (torso)

I think it is ok to merge, travis is also fine :)
@lornat75 @randaz81, what do you think?

lornat75:
Fine with me to merge.

randaz81:
already checked and approved :-)

2015-01-12 11:58 GMT+01:00 Lorenzo Natale <notifications@github.com>:

pattacini:
@barbalberto can we delete the old branch then?

barbalberto:
Done, thanks. I forgot. :)

pattacini:
@barbalberto, can we fix the following 3 warnings? They showed up after the merge:

```
(1) icub-main\src\modules\skinManager\src\SkinDiagnosticsReadThread.cpp(110): warning C4244: 'initializing' : conversion from 'double' to 'int', possible loss of data
(2) icub-main\src\modules\skinManager\src\SkinDiagnosticsReadThread.cpp(118): warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)
 skinManager.cpp
(3) icub-main\src\modules\skinmanager\src\skinmanager.cpp(109): warning C4715: 'iCub::skinManager::skinManager::configure' : not all control paths return a value
skinManager.vcxproj -> C:\dev\icub-main\build\bin\Release\skinManager.exe
```

Especially (2) and (3) can be dangerous...

For (2) I'd replace `bool = int & int;` with `bool = (int!=0) && (int!=0);` unless the original goal was to actually perform a bit-mapping operation, then `bool = (int & int)!=0;`

barbalberto:
I guess (2) is converting a **int with bitwise** error code into a **vector of bool** by comparing the integer with a 1 shifting its position each loop.
also  errorMask = errorMask * 2,  imho could  be   
errorMask = errorMask << 1;  (bitwise meaning is more clear to read)

Am I right @JoErNanO ? Is this the intended behaviour?

(3) fixed by adding a missing return true

PR #104, opened 2015-02-12
Added virtual skin capabilities to iCubSkinGuiQT

alecive:
In sight of improving the periPersonalSpace demo, I have added the capability for `iCubSkinGui` to process both a virtual and a real contact (and I also did minor fixes over the time). This is how it works:

 * The previous API is intact
 * A new port is opened (instead of `/skinGui/part:i` now there is also `/skinGui/part_virtual:i`).
 * The parsing of the new `virtual` port is slightly different: instead of simply a list of vectors, the user can specify a new color as well
 * The real skin has priority over the virtual one, with the virtual skin is processed only if the real skin is under a certain threshold (tunable with the parameter `--skinThreshold`, default 15.0)
 * The logic is triangle-based: that is, it may be that a triangle gets a big enough real contact, whereas the next one gets activated by a virtual one.


pattacini:
:+1:

PR #120, opened 2015-03-11
iCub::ctrl::{axis2dcm, dcm2euler, ...} moved into yarp::math

pattacini:
This change might break compilation of external modules/repos.
`icub-main` has been updated, though


pattacini:
This PR follows up https://github.com/robotology/yarp/pull/437

PR #126, opened 2015-03-17
New torque control law

randaz81:
added new configuration parameters for the torque controller:
in section [TORQUE_CONTROL]
controlLaw allows to select a specific torque control law.
controlUnits allows to specify the measurement units for the torque controller gains/params.
if [TORQUE_CONTROL] section is not found or if an invalid controlLaw or controlUnits value is detected, torque control will be disabled (ETH robots only).
Old CAN robot can still use [TRQ_PIDS] section, but will be deprecated soon.


barbalberto:
Just a comment
Why deprecate the TRQ_PID and then put the same information inside TORQUE_CONTROL? It looks more confusing to me, unless then we (you) want to do the same for position pids

Can't have both? Use TRQ_PID as usual and the new TORQUE_CONTROL group for controlLaw?
I guess the important difference is the measurement units that can change (dangerous to let user choose the meas unit from congif file imho), but if TORQUE_CONTROL is missing you can default to old behaviour (hardware_units, I suppose)

PR #130, opened 2015-03-27
Position move time check

randaz81:


src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
```diff
@@ -2347,6 +2349,13 @@ bool embObjMotionControl::setPositionModeRaw()
 
 bool embObjMotionControl::positionMoveRaw(int j, double ref)
 {
+    if (_last_position_move_time[j] == 0) _last_position_move_time[j] = yarp::os::Time::now()-1.0;
```
lornat75: Why this?

lornat75:
Please wait some time before merging pull requests that are non-trivial like this one!

randaz81:
Well, actually it's trivial... It's just a warning message. Before merging, it was extensively tested on both CAN and ETH architecture. RobotInterfaces now prints a warning (but still accepts the command) if the user misuses positionMove(). (the time interval between two position commands should be >80ms)

randaz81:
The highlighted line of code was removed in master. I was not able to push the commit before. Today github server has some network problems.

PR #145, opened 2015-04-30
updated ETH robot configuration files to use metric units in position control

randaz81:
affected robots:
iCubGenova02
iCubGenova04
iCubDarmstadt01
iCubSheffield01
iCubHeidelberg01


randaz81:
Tested on iCubGenova02 & iCubGenova04

PR #167, opened 2015-06-30
Do not copy yarprun / yarpserver in /tmp

Tobias-Fischer:
Fix for https://github.com/robotology/icub-main/issues/154

Works fine on our setup, and has the advantage that `LD_LIBRARY_PATH` does not need to be set, if `INSTALL_WITH_RPATH` is used.

Best, Tobias


pattacini:
thanks @Tobias-Fischer

PR #177, opened 2015-07-16
Added --ros flag for running yarpserver remotely in the cluster manager scripts

plinioMoreno:
* Added a checkbox in the cluster gui (icub-cluster.py)
* Added a third parameter in the bash script that now deals with the multiple options (icub-cluster-server.sh)

Let me know if you prefer to create new files or agree with my pull request

### DIFF
--- a/app/iCubCluster/scripts/icub-cluster.py
+++ b/app/iCubCluster/scripts/icub-cluster.py
@@ -176,6 +176,12 @@ class App:
         self.bCheckNs.config(state=DISABLED,disabledforeground="#00A000")
         self.bCheckNs.grid(row=1, column=1, sticky=W)
 
+	self.bROSoption = Entry(tmpFrame)
+	self.ROSoption=IntVar()
+	self.ROSoption.set(0)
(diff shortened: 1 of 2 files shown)


app/iCubCluster/scripts/icub-cluster.py
```diff
@@ -176,6 +176,12 @@ def setCluster(self, cluster):
         self.bCheckNs.config(state=DISABLED,disabledforeground="#00A000")
         self.bCheckNs.grid(row=1, column=1, sticky=W)
 
+	self.bROSoption = Entry(tmpFrame)
+	self.ROSoption=IntVar()
+	self.ROSoption.set(0)
+	self.bROSoption=Checkbutton(tmpFrame, variable=self.ROSoption)
+        self.bROSoption.grid(row=1, column=2)
```
lornat75: can you double check this identation does not cause python to get angry?

lornat75:
This script is still using yarperver3 which is now default (yarpserver is actually calling yarpserver3). Can you fix this since you are at it? Thanks :)

plinioMoreno:
No problem, I'll change the script anc create a new pull request

PR #178, opened 2015-07-17
Added --ros flag for running yarpserver remotely in the cluster manager scripts

plinioMoreno:
* icub-cluster.py -- now has the --ros option and does not check for yarpserver3
* icub-cluster-server.sh -- the yarpserver3 cases are removed and ros option is available


plinioMoreno:
@lornat75, just a reminder to merge.

PR #183, opened 2015-08-06
Usb camera driver

barbalberto:
This branch implements the yarp driver for USB camera. It is meant to work with SeeCam cameras for CER but it should work also with a more wide range of cameras as long as they publish data in a standard format.

Done:
- query of available formats and capabilities from the camera
- acquisition with any supported width and height 
- conversion in rgb
- camera controls (brightness, saturation ...) except Shutter config. They can be set from terminal using yarp rpc /grabber

To be done
- update frameGrabberGui2
- shutter config
- find unique global id for usb device (to discriminate between left/right)

@drdanz: can you check the FindV4L2.cmake


drdanz:
@barbalberto Why is it in the icub-main repository? I thought this was supposed to be in the YARP repository (as for the firewire devices).

About the FindV4L2.cmake file, debian/ubuntu packages ship pkg-config files:

```
$ pkg-config --list-all | grep v4l
libv4l2                             libv4l2 - v4l2 device access library
libv4l1                             libv4l1 - v4l1 compatibility library
libv4lconvert                       libv4lconvert - v4l format conversion library
libv4l2rds                          libv4l2rds - v4l2 RDS decode library
```

You should use pkg-config when available, that means that instead of your custom stuff you can use MacroStandardFindModule, have a look at the other cmake files in yarp (for example [FindGooCanvas.cmake](https://github.com/robotology/yarp/blob/master/conf/FindGooCanvas.cmake))

barbalberto:
Well, the dragonly (firewire cameras) are in icub repo, so I put them there as well but, yes I agree they will fit better in the yarp repo.

Thanks for the cmake tip  ;-)  I'll do it.

drdanz:
Oh sorry, I thought it was in yarp...

barbalberto:
I implemented the changes you suggested.
Can someone give a feedback about win & mac?

Right now the cmake should fail when you activate the device with an error saying the OS is not supported yet.

PR #197, opened 2015-09-17
Add yarp motion control class for CER robot tripod device.

barbalberto:
It implements bacis motion control interfaces and calls an Inverse Kinematic
library (fake for now) to converting new references from user space to HW space
and converting encoder reading from HW to user.

It then can be attached to a lower level device (like can/embObjMotionControl)
to send commands to real HW devices. It can do it also remotely by mean of
a remoteControlBoard.


pattacini:
Agreed with @barbalberto that we should have a dedicated repo for `CER`...
Closing

PR #206, opened 2015-10-08
added support for xi:include

randaz81:
@drdanz 
added support for xi:include directive

### DIFF
--- a/app/robots/iCubGenova02_xi/CMakeLists.txt
+++ b/app/robots/iCubGenova02_xi/CMakeLists.txt
@@ -1,0 +1,11 @@
+set(appname iCubGenova02)
+
+set(scripts icub_all.xml icub_all_inertials.xml robotInterface.ini general.xml)
+
+yarp_install(FILES ${scripts} DESTINATION ${ICUB_ROBOTS_INSTALL_DIR}/${appname})
+yarp_install(DIRECTORY calibrators DESTINATION ${ICUB_ROBOTS_INSTALL_DIR}/${appname} PATTERN .svn EXCLUDE)  
+yarp_install(DIRECTORY cartesian DESTINATION ${ICUB_ROBOTS_INSTALL_DIR}/${appname} PATTERN .svn EXCLUDE)
+yarp_install(DIRECTORY wrappers DESTINATION ${ICUB_ROBOTS_INSTALL_DIR}/${appname} PATTERN .svn EXCLUDE)
+yarp_install(DIRECTORY hardware DESTINATION ${ICUB_ROBOTS_INSTALL_DIR}/${appname} PATTERN .svn EXCLUDE) 
+
+
--- a/app/robots/iCubGenova02_xi/calibrators/head-calib.xml
+++ b/app/robots/iCubGenova02_xi/calibrators/head-calib.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE devices PUBLIC "-//YARP//DTD robotInterface 1.0//EN" "http://www.icub.org/DTD/robotInterfaceV1.0.dtd">
+
+<devices robot="iCubGenova02" build="1">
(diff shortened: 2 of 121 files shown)


randaz81:
work in progress

barbalberto:
I wounldn't call it "support for xi:include" :-P 
it is more like 
change the string "param file" to the string "xi:include" 
This will not make it w3c standard compliant
:-P :-P :-P

drdanz:
I almost had a stroke when I saw "Files changed **121**" :sweat_smile:

This is a very basic support of the XInclude specification (support for the other attributes and fallback), but for our needs it looks good, but I have a few comments:

There is another small issue in the xml files: `xi:include` is a bit wrong if the namespace `xi` is not defined somewhere, and you defined it only in the main xml, that means (I think) that devices and params files using xi:include (for example `app/robots/iCubGenova02_xi/calibrators/head-calib.xml`) are not well formatted xml files. I suggest you add the definition to the "devices" and "params" files as well. I'm not sure if this could cause any issue, but perhaps some tool could complain.

I wasn't sure about the `itype` attribute, but I read on the [specification](http://www.w3.org/TR/xinclude/#include_element) that
> Attributes other than those listed above may be placed on the xi:include element.

and the xinclude schema contains

```xml
<xs:element name="include" type="xi:includeType" />
<xs:complexType name="includeType" mixed="true">
  <!-- ... -->
  <xs:anyAttribute namespace="##other" processContents="lax"/>
</xs:complexType>
```


That means that it is ok to use custom attributes. Anyway


```xml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE robot PUBLIC "-//YARP//DTD robotInterface 1.0//EN" "http://www.icub.org/DTD/robotInterfaceV1.0.dtd">

<robot name="theRobot"
       build="1"
       portprefix="icub"
       xmlns:xi="http://www.w3.org/2001/XInclude">
  <xi:include href="foo.xml" itype ="params"/>
</robot>
```
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE params PUBLIC "-//YARP//DTD robotInterface 1.0//EN" "http://www.icub.org/DTD/robotInterfaceV1.0.dtd">

<params robot="theRobot" build="1">
   <param name="foo">bar</param> 
</params>
```

When processed will generate this xml:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE robot PUBLIC "-//YARP//DTD robotInterface 1.0//EN" "http://www.icub.org/DTD/robotInterfaceV1.0.dtd">

<robot name="theRobot" build="1"
       portprefix="icub"
       xmlns:xi="http://www.w3.org/2001/XInclude">
  <params robot="theRobot" build="1">
    <param name="foo">bar</param> 
  </params>
</robot>
```

that is different from this (i.e. the the file without the inclusion)

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE robot PUBLIC "-//YARP//DTD robotInterface 1.0//EN" "http://www.icub.org/DTD/robotInterfaceV1.0.dtd">

<robot name="theRobot" build="1"
       portprefix="icub"
       xmlns:xi="http://www.w3.org/2001/XInclude">
  <param name="foo">bar</param> 
</robot>
```

And I think that query results will change (and it might have a big influence when you want to mass change the xml for all the robots at the same time.
Therefore I suggest you to investigate if the `xinclude` attribute could be used instead of `itype`.

For example
```xml
  <xi:include href="foo.xml" parse="xml" xpointer="xpointer(//params/*)"/>
```

should do the trick and work as if the extra `<params>` tag was not there

randaz81:
I'm sorry, I wanted to delete this pull request, but I was able to find only a delete branch options (not good!) or close button. These commits are not updated, so please do not consider them yet. I used this branch only for internal testing and many of your considerations have been already taken in account.
btw: xpointer was considered but not possibile in our case. I ended up by removing all params and devices sections. (>3000 files :) )

PR #239, opened 2015-11-12
fixing configuration files for iCubDarmstadt01

serena-ivaldi:
3 files changed


pattacini:
Two files could live on your local repo without the need for being pushed here. I'm referring to `general_debug_risky.xml` and `icub_all_without_skin_legs.xml`. They represent indeed temporary workaround.

Regarding the third file, there are weird comments leading to think it doesn't contain the final values yet.

I'm for rejecting this PR.

serena-ivaldi:
The calibration file contains the final values, but I kept the previous ones on commented lines. This has to be taken! 

The general_debug_risky.xml is for taking the raw values. We didn't have this - why rejecting it? 

As for the other file it is necessary since iCubDarmstadt has apparently an hardware issue on the skin of the legs. It is not a temporary workaround: if I understood correctly they have had this problem for a while and there is no plan to fix the legs skin. 
I think you should add it to the repository (only for iCubDarmstadt) otherwise the next time they git clone or pull and then install, they will wipe the new files and will experience a lot of errors in the robotInterface as they were doing before my fix.

robertocalandra:
@pattacini I think it would make a lot of sense to accept this pull request. the changes are specific to the iCubDarmstadt01 and would not effect any other user. I don't see any other valid reason why you should refuse the request.

pattacini:
@randaz81 is the chief maintainer of robots apps, so he will decide how to proceed.

In my view, no robot has either `general_debug_risky.xml` or `icub_all_without_skin_legs.xml`; even the name of those files suggest they're for a temporary workaround. Problems need to be addressed upstream.

Further, we tend to keep all robots directory as similar as possible with respect to their `xml` files, and soon there'll be a new automatic system to handle them.

Concerning the wrist, I'd suggest to keep only the new values then. We rely on git to access the history. In this respect, I don't know what is the current story about updating the excel file accordingly, that is, if it still needs to be done or not.

robertocalandra:
Ok, how do you suggest to address the problem upstream?

pattacini:
@robertocalandra it's not because there's the risk to affect other robots. We are the maintainers and we have policies for how to manage this repository.

I've just expressed my opinion; however, as said, @randaz81 will deal with this PR.

pattacini:
A longstanding issue with the legs skin suggests that the problem should be tackled hardware-wise with high priority.

randaz81:
Rejected. I'll do the fix on your files.

randaz81:
fix pushed into master.

Issue #235, opened 2015-11-04
Skin ports remain up after closing robotInterface


DanielePucci:
I am currently working on Purpley (`iCubGenova02`), and after closing `robotInterface` some ports related to the skin stay up.

Feel free to close this issue if it is a duplicate of some other issue.

![screenshot from 2015-11-04 10 13 23](https://cloud.githubusercontent.com/assets/5045846/10934085/bb977b3a-82dd-11e5-9863-5e3a35632973.png)

@marcoaccame @pattacini @drdanz @francesco-romano

DanielePucci:
@barbalberto

pattacini:
Another issue which seems suitable for young researchers. It would be helpful to exercise their debugging abilities and also for them to get more feeling with the gut of our system.

@tanismar @massimoregoli @GiuliaP @bhigy @giuliavezzani @jeljaik and whoever wants to be remembered as the quest's winner.

pattacini:
@bhigy would you be willing to take up this task?

bhigy:
I'll try to have a look when I find a moment.

pattacini:
Great, thanks.

lornat75:
The first thing I would check is if ports are created dynamically or statically. If they are created dynamically (i.e. using new) check that close (and destructors) call delete.

giuliavezzani:
If @bhigy is too busy in this period, I can try to fix the problem the next
month,  as soon as all the bureaucratic tasks will be finished :)

2015-11-22 11:04 GMT+01:00 Lorenzo Natale <notifications@github.com>:

barbalberto:
If I remember correctly, the skin wrapper create dinamically a set of analogServers using ``new``.
Maybe something is not right in the cleanup.
Both classes are located in the icub-main/icubmod, if someone wants to address this issue I can help giving a couple of hints for a quick start.

PR #250, opened 2015-11-26
fixed bug on Skin ports issue #235

giuliavezzani:
Introduced changes:
- closing skin ports that were remaining open : close function is now called inside threadRelease function
- threadRelease is  now called  only by RateThread, whereas before it was called twice

### DIFF
--- a/src/libraries/icubmod/analogServer/analogServer.cpp
+++ b/src/libraries/icubmod/analogServer/analogServer.cpp
@@ -181,6 +181,7 @@ AnalogServer::AnalogServer(): RateThread(0)
 AnalogServer::~AnalogServer()
 {
     yTrace();
+    close();
     _rate = 0;
     analogSensor_p = NULL;
 }
@@ -414,7 +415,6 @@ void AnalogServer::threadRelease()
     {
         analogPorts[i].port.close();
     }
-    close();
 }
 
 void AnalogServer::run()
--- a/src/libraries/icubmod/analogServer/analogServer.cpp
+++ b/src/libraries/icubmod/analogServer/analogServer.cpp
@@ -181,7 +181,6 @@ AnalogServer::AnalogServer(): RateThread(0)
 AnalogServer::~AnalogServer()
 {
     yTrace();
-    threadRelease();
     _rate = 0;
     analogSensor_p = NULL;
 }
@@ -415,6 +414,7 @@ void AnalogServer::threadRelease()
     {
         analogPorts[i].port.close();
     }
+    close();
 }
 
 void AnalogServer::run()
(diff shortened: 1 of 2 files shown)


pattacini:
The `close()` service needs to be called from outside the driver (as the `open()` method). The function `close()` does contain the call to `stop()` that in turn will call the `threadRelease()`. Thus, there's still something strange to be investigated.

bhigy:
Would it be good if we call it from skinWrapper::close() (analogServer is belonging to skinWrapper)?

barbalberto:
I see. Since the ``SkinWrapper`` directly calls the destructor, then the ``close()`` function was never called.

Can you check one thing please? Is the ``threadRelease()`` or ``close()`` called more then once by enabling env var YARP_TRACE_ENABLE=1?

Now that we (you) found the point, I believe the cleanest way is to have the destructor af analogServer simply call its own close function and move all the cleaning stuff there. maybe removing the explicit calls to threadRelease if it is redundant (it should be called by stop)

pattacini:
In case the destructor will call the `close()`, we then make sure that calling `close()` twice in a row won't bring about any problem. This is because `close()` can be called also from outside the driver.

barbalberto:
yes, that's true. SkinWrapper should definitely calls the close before the destructor.
Usually the desctructor should take care of cleaning things regardless of what happened in the past and not relyng on user code (close function called or not)

barbalberto:
@pattacini yes, that's the correct behaviour. Close function can always be called more then once by the user so it must be safe in any case.

bhigy:
One question regarding AnalogServer. The source file is in icub-main but it is part of yarp::dev namespace. Shouldn't it be in yarp repo?

pattacini:
The inclusion of classes in `yarp::dev` outside `yarp` is at least misleading and confusing. It might be a leftover of the past and I cannot honestly foresee what will be the effect of removing that inclusion. Also, it is not the only driver doing so (see for example the `parametricCalibrator`).

Usually, the implementation of iCub specific drivers is located in `icub-main` repo.

Good catch anyways @bhigy

barbalberto:
Most of the devices in icubmod are in the yarp::dev namespace. It's most probably a copy & paste leftover. One day we will fix it ... maybe next IFI day? :smile:

giuliavezzani:
@bhigy  and me are thinking to put the close( ) of AnalogServer in its destructor to solve the problem. Is it necessary also to call it in the SkinWrapper, before the delete?
We also tried to call the close() function twice and it was ok.

barbalberto:
Great work @bhigy and @giuliavezzani !!
Yes, I'll put the close also in the SkinWrapper, as good practise.

pattacini:
Then, does everybody agree on merging this PR straightaway?

PR #258, opened 2015-12-08
ctrlLib inherits math constants from cmath and publish them with cmake

pattacini:
As discussed in https://github.com/robotology/QA/issues/81, **`ctrlLib`** now depends on **`cmath`** to import mathematical constants (e.g. `M_PI`) instead of re-defining them from scratch.

This dependency is then published by means of suitable cmake macro for systems using **MSVC** compilers. In this respect, I have explicitly avoided referring to corresponding Cmake versions in order not to retain some unnecessary leftover in Windows (as said, Windows users have more recent versions of Cmake). Let's see if this will suffice.

I've also given a go to the compilation of `wysiwyd`, `karma` and `iol` sister repositories without any problem related to this specific change.

/cc @traversaro

### DIFF
--- a/src/libraries/ctrlLib/CMakeLists.txt
+++ b/src/libraries/ctrlLib/CMakeLists.txt
@@ -14,7 +14,7 @@ set(folder_source src/math.cpp
                   src/functionEncoder.cpp
                   src/optimalControl.cpp
                   src/neuralNetworks.cpp
-				          src/outliersDetection.cpp)
+                  src/outliersDetection.cpp)
 
 set(folder_header include/iCub/ctrl/math.h
                   include/iCub/ctrl/filters.h
@@ -26,7 +26,7 @@ set(folder_header include/iCub/ctrl/math.h
                   include/iCub/ctrl/functionEncoder.h
                   include/iCub/ctrl/optimalControl.h
                   include/iCub/ctrl/neuralNetworks.h
-				          include/iCub/ctrl/outliersDetection.h)
+                  include/iCub/ctrl/outliersDetection.h)
 
 source_group("Source Files" FILES ${folder_source})
 source_group("Header Files" FILES ${folder_header})
--- a/src/libraries/ctrlLib/CMakeLists.txt
+++ b/src/libraries/ctrlLib/CMakeLists.txt
@@ -2,11 +2,9 @@
 # Authors: Ugo Pattacini
 # CopyPolicy: Released under the terms of the GNU GPL v2.0.
 
-SET(PROJECTNAME ctrlLib)
+project(ctrlLib)
 
-PROJECT(${PROJECTNAME})
-
-SET(folder_source src/math.cpp
+set(folder_source src/math.cpp
                   src/filters.cpp
                   src/kalman.cpp
                   src/pids.cpp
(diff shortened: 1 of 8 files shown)


lornat75:
I still have to understand the cmake code, but I do not like the if(msvc)..,

If we address this problem I would consider removing any definition from ctrllib public header files.

On 08/dic/2015, at 12:38, "Ugo Pattacini" <notifications@github.com<mailto:notifications@github.com>> wrote:


As discussed in robotology/QA#81<https://github.com/robotology/QA/issues/81>, ctrlLib now depends on cmath to import mathematical constants (e.g. M_PI) instead of re-defining them from scratch.

This dependency is then published by means of suitable cmake macro for systems using MSVC compilers. In this respect, I have explicitly avoided referring to corresponding Cmake versions in order not to retain some unnecessary leftover in Windows (as said, Windows users have more recent versions of Cmake). Let's see if this will suffice.

I've also given a go to the compilation of wysiwyd, karma and iol sister repositories without any problem related to this specific change.

/cc @traversaro<https://github.com/traversaro>

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/robotology/icub-main/pull/258

Commit Summary

  *   [ctrlLib]: M_PI is now inherited from cmath and this dependency is published using suitable cmake macro

File Changes

  *   M src/libraries/ctrlLib/CMakeLists.txt<https://github.com/robotology/icub-main/pull/258/files#diff-0> (35)
  *   M src/libraries/ctrlLib/include/iCub/ctrl/math.h<https://github.com/robotology/icub-main/pull/258/files#diff-1> (7)
  *   M src/libraries/iKin/CMakeLists.txt<https://github.com/robotology/icub-main/pull/258/files#diff-2> (19)
  *   M src/modules/camCalibWithPose/CMakeLists.txt<https://github.com/robotology/icub-main/pull/258/files#diff-3> (5)
  *   M src/simulators/iCubSimulation/CMakeLists.txt<https://github.com/robotology/icub-main/pull/258/files#diff-4> (4)
  *   M src/simulators/iCubSimulation/wrapper/iCubSimulationControl.cpp<https://github.com/robotology/icub-main/pull/258/files#diff-5> (1)
  *   M src/tools/iCubGui/src/CMakeLists.txt<https://github.com/robotology/icub-main/pull/258/files#diff-6> (3)
  *   M src/tools/iCubGui/src/mesh.h<https://github.com/robotology/icub-main/pull/258/files#diff-7> (1)

Patch Links:

  *   https://github.com/robotology/icub-main/pull/258.patch
  *   https://github.com/robotology/icub-main/pull/258.diff

pattacini:
The define `_USE_MATH_DEFINES` is meaningful only for MSVC compilers, that was the reason why.

We have `M_PI` used for default parameters values in the public headers of e.g. `iKin`; this prevent us from removing them.

lornat75:
...at the cost of sounding like a broken record... iKin should not depend on these symbols, this will cause troubles. you have already introduced changes that require modifications in cmake code, why don't you put these symbols in a namespace and modify iKin headers accordingly?

pattacini:
Perhaps I'm missing something or I might be wrong but `M_PI` is a `#define`, hence it's not an exported symbol and as such it cannot be even confined within a namespace. It's only a number in the end and I need it for defining default behaviors of some methods in the API. The use of `M_PI` is mandatory for me, unfortunately.

The only change I've introduced applies to a couple of `CMakeLists.txt` files upstream without affecting all the dependent code, which was my original main concern. As far as I've tested, compilation runs smoothly on Windows, the only platform touched by this change, because it's the only platform requiring the definition of `_USE_MATH_DEFINES` to get access to `M_PI` (if I'm correct).

The improvement is that we don't re-define `M_PI` in our code and we get rid of the occasional warnings popping out here and there.

PR #259, opened 2015-12-11
ctrlLib defines math symbols to avoid referring to M_PI in the headers

pattacini:
Following up discussions in https://github.com/robotology/QA/issues/81, `ctrlLib` declares now math symbols within its own namespace to avoid referring explicitly to e.g. `M_PI` in the public headers.

To get `icub-main` compiled I had to apply changes here and there. I'm afraid that this new approach will have an impact on sister projects: I think it is no longer guaranteed that they'll compile smoothly.


pattacini:
I'm going to check sister projects compilation.

PR #261, opened 2015-12-14
added new "get reference" methods 

randaz81:
@barbalberto 

3x getRefPositionRaw()
returns the last setPosition (from directPosition)

3x getTargetPositionRaw()
returns the last positionMove (from iPosition)

3x getRefVelocityRaw()
returns the last velocityMove (from iVelocity. Maybe getTargetVelocityRaw is a better name?)


barbalberto:
I think it is good getRefVelocity as you choose, otherwise it seems that getTargetPosition/Velocity are related, which is not true.

PR #313, opened 2016-04-01
splittedanalogdevices

marcoaccame:
There are three new devices in icubmod: embObjMais, embObjStrain and embObjInertials. They are disabled by default and not yet used in robotInterface. They come from the splitting of embObjAnalog. This split allows to make embObjInertials evolve to better support new sensors. Also, they are the basis for reading from xml files the properties of MAIS, STRAIN and MTB can boards and sending them to the relevant ETH boards for their runtime config.


randaz81:
@marcoaccame 
some notes about the xml files:

* I'd remove params:
             
```
<group name="CANBOARDS">
                    <param name="type">                 enumCanBrdSTRAIN    </param>
<group name="SENSORS">
                    <param name="type">                 enumSnsrSTRAIN       </param>
```
They can be automatically inferred and set inside the plugin, since each plugin can only work with a specific type of sensor (e.g. embObjInertial can only work with type enumCanBrdMTB, embObjStrain can only work with type enumCanBrdSTRAIN).

* I'd prefer to have the following values hard-coded in the plugin and not-robot specific.
```                    
<group name="PROTOCOL">
                        <param name="major">            1                   </param>    
                        <param name="minor">            0                   </param>     
</group>                    
<group name="FIRMWARE">
                        <param name="major">            0                   </param>    
                        <param name="minor">            0                   </param> 
                        <param name="build">            0                   </param>
</group>
```
If these params are hard-coded in the device plugin, when a new feature (which breaks the compatibility) is introduced, the developer just need to update the numbers in the code. Then, if another user updates icub-main without upgrading the firmware, robotInterface will close displaying a "please update the fw" error message.
On the contrary, if these numbers are written in the xml files, when a new feature is introduced, the configuration files of all the robots needs to updated (a tedious and error-prone work).
The legacy `canmotioncontrol` plugin has the protocol requirements hard-coded in `icub-main\src\libraries\icubmod\motionControlLib\messages.h` and this paradigm worked well so far without any complain.

PR #358, opened 2016-07-11
Enhanced loader tools

marcoaccame:
This is an enhancement to canLoader and ethLoader which now give the user more information about the running CAN and ETH boards.

It is an intermediate step towards a merging of the two GUIs into a single one which uses the Qt graphical library and adds some new features.

The GUIs in this branch have some enhancements in the graphical layout and also have some more information.

The new canLoader and ethLoader work with existing FW but some of the additional information are available only if the ETH boards 
are updated with new FW for the loader, updater, and application partitions (see icub-firmware-build/EMS/howto.txt for details about 
the partitions in an ETH board).

The main change in canLoader is that it allows to discover at the same time CAN boards attached to both CAN1 and CAN2 of ETH boards.
It also adds information about the executing partition (CAN bootloader or application) and about the running CAN protocol version.
If the ETH boards run the new application FW, then the user does not need to run a discovery with ethLoader to use the canLoder.

The main change in ethLoader is that it prints the type of ETH board, the executing partition with related version, date, built date, 
it manages a descriptive string of 30 bytes stored in the EEPROM of the ETH board, it protects vs accidental deletion of the EEPROM.
The new info/features are available with the new FW only.
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/robotology/icub-main/pull/358?approve=1'></a>&nbsp;<a href='https://github.com/robotology/icub-main/pull/358'></a>
<a href='#crh-end'></a>


drdanz:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/robotology/icub-main/pull/358'>Refresh</a>.<a href='#crh-update-none'></a>

PR #365, opened 2016-08-30
runtime configuration of motion control

marcoaccame:
This change applies to all ETH robots and adds to embObjMotionControl the capability to retrieve from 
an xml file the HW configuration of sensors and actuators managed by the ETH board (EMS4 etc.).

If the device does not find a proper SERVICE group or if SERVICE.type = eomn_serv_MC_generic,
the embObjMotionControl does not send the configuration and the ETH board works as
it has done so far: it configures itself according to hard-coded settings selected by its IP address.

Otherwise, if the device finds a valid configuration it sends it to the ETH board which uses it to 
initialize all it needs to perform motion control: type and location of sensors and actuators plus
related control parameters.

The system has been extensively tested on iCubGenov02 and on the new V3 head.
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/robotology/icub-main/pull/365?approve=1'></a>&nbsp;<a href='https://github.com/robotology/icub-main/pull/365'></a>
<a href='#crh-end'></a>


drdanz:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/robotology/icub-main/pull/365'>Refresh</a>.<a href='#crh-update-none'></a>

PR #393, opened 2016-12-02
Fix random crash into ParametricCalibrator

barbalberto:
This is a quick fix for what it looks like a memory problem ... to be further investigated


barbalberto:
Tested on the robot.
Merging because it is required on iCub Black

PR #395, opened 2016-12-07
iCub_Sim: fix error in checkMotionDone

barbalberto:
Fix common error in `checkMotionDone`.
That function returns a single value even when asked for more than one joint.

### DIFF
--- a/src/simulators/iCubSimulation/wrapper/iCubSimulationControl.cpp
+++ b/src/simulators/iCubSimulation/wrapper/iCubSimulationControl.cpp
@@ -928,14 +928,12 @@ bool iCubSimulationControl::checkMotionDoneRaw (bool *ret)
     _mutex.wait();
     bool fin = true;
     for(int axis = 0;axis<njoints;axis++)
-        {
-            if(! (fabs( current_jnt_pos[axis]-next_pos[axis])<error_tol[axis]))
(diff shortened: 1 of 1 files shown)


src/simulators/iCubSimulation/wrapper/iCubSimulationControl.cpp
```diff
@@ -1162,7 +1160,7 @@ bool iCubSimulationControl::checkMotionDoneRaw(int nj, const int * jnts, bool *d
     bool ret = true;
     for (int i = 0; i<nj; i++)
     {
-        ret &= checkMotionDoneRaw(jnts[i], &done[i]);
+        ret &= checkMotionDoneRaw(jnts[i], done);
```
pattacini: Here we should "accumulate" rather than overwriting, something like this:

```cpp
if (done==NULL)
    return false;

bool ret=true;
*done=true;
for (int i=0; i<nj; i++)
{
    bool tmpDone;
    ret&=checkMotionDone(jnts[i],&tmpDone);
    *done&=tmpDone;
    if (!*done)
        break;
}
```

barbalberto:
Oooppss!!
Yes, you are right!!

barbalberto:
Updated

PR #398, opened 2016-12-13
Better handling of homing gaze in ARE

pattacini:
The way **actionsRenderingEngine** (ARE) is currently homing the gaze is somewhat mazy. The `goHome()` method triggers homing by changing the state of a internal variable which in turn is handled within `run()`.

Instead, this PR brings everything inside `goHome()`.


pattacini:
Importantly, this PR solves weird homing of the gaze people have been seeing at Imperial, UPF, Sheffield, although everything seems fine at IIT.

PR #399, opened 2016-12-14
Fwupd enhanced

marcoaccame:
Here are some enhancements of the original GUI.
1. reordered layout of columns in left panel.
2. reorganised buttons and added tooltips.
2. tool buttons for completing selection and deselection of all boards.
3. properties on right panel are printed also for can boards.


marcoaccame:
I opened the wrong PR, hence I closed it.
The correct one is: https://github.com/robotology/icub-main/pull/400

PR #410, opened 2017-01-18
V4L_camera: added the inheritance from yarp::dev::IRgbVisualParams and implemented the relative virtual methods.

Nicogene:
Please review code.

### DIFF
--- a/src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
+++ b/src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
@@ -90,6 +90,13 @@ V4L_camera::V4L_camera() : RateThread(1000/DEFAULT_FRAMERATE), doCropping(false)
     param.camModel = SEE3CAMCU50;
     myCounter = 0;
     timeTot = 0;
+    configFx = true;
+    configFy = true;
+    configPPx = true;
+    configPPy =true;
+    configRet = true;
+    configDistM = true;
+    configIntrins = true;
 }
 
 yarp::os::Stamp V4L_camera::getLastInputStamp()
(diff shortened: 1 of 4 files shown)


src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -102,6 +109,65 @@ yarp::os::Stamp V4L_camera::getLastInputStamp()
...
+    else if(bt.find("name").isNull() || bt.find("k1").isNull()
+            || bt.find("k2").isNull() || bt.find("k3").isNull()
+            || bt.find("t1").isNull() || bt.find("t2").isNull())
+        configIntrins=false;
+
+    param.width = config.check("width",yarp::os::Value(320),
+                     "desired width of test image").asInt();
+    param.height = config.check("height",yarp::os::Value(240),
+                     "desired height of test image").asInt();
+    param.horizontalFov=config.check("horizontalFov",Value(0.0),
```
barbalberto: `width` and `hight` parameters are already checked in the `fromConfig` function (line 165-272).
That function was meant to collect all the config file parsing into a single place, so I'd suggest to move all the addictional `config.check()` there and remove the double check on `width` and `hight`

src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -102,6 +109,65 @@ yarp::os::Stamp V4L_camera::getLastInputStamp()
  */
 bool V4L_camera::open(yarp::os::Searchable& config)
 {
+    yarp::os::Value *val;
+    Value* retM;
+    retM=Value::makeList("1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 1.0");
+    yDebug()<<"Checking the configuration file";
```
barbalberto: Remove the debug print

src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -90,6 +90,13 @@ V4L_camera::V4L_camera() : RateThread(1000/DEFAULT_FRAMERATE), doCropping(false)
     param.camModel = SEE3CAMCU50;
     myCounter = 0;
     timeTot = 0;
+    configFx = true;
```
barbalberto: I'd initialize it to false and set the flag to true when the param is actually found.
_The result is the same_ as long as the code is bug free, but it looks more safe to me in case any bug will be introduced in the future.
The idea is : _assuming theere is a bug somewhere_, it is safer a device that says "I don't have this parameter" when the parameter is correctly loaded, then a device that says "I have this parameter" when it actually don't.

src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -160,6 +226,47 @@ bool V4L_camera::open(yarp::os::Searchable& config)
...
+bool V4L_camera::getRgbFOV(double &horizontalFov, double &verticalFov){
+    horizontalFov=param.horizontalFov;
+    verticalFov=param.verticalFov;
+    return configFx && configFy;
+}
+
+bool V4L_camera::setRgbFOV(double horizontalFov, double verticalFov){
+    param.horizontalFov=horizontalFov;
+    param.verticalFov=verticalFov;
+    return true;
```
barbalberto: Ops, I forgot to say this method shall actually change the resolution of the video streaming, so it should call the `deviceUninit` function to stop the streaming, then change the param setting and then restart the streaming video with 'deviceInit'.

src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -160,6 +226,47 @@ bool V4L_camera::open(yarp::os::Searchable& config)
...
+    return configIntrins;
+}
+
+bool V4L_camera::getRgbMirroring(bool &mirror){
+    mirror=param.mirror;
+    return true;}
+
+bool V4L_camera::setRgbMirroring(bool mirror){
+    param.mirror=mirror;
+    return true;
```
barbalberto: This method shall try to set the HW mirroring option of the camera, if available.

PR #412, opened 2017-01-19
Implementation of runtime resolution change for usbCamera (on Linux)

Nicogene:
Please review code.


src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -694,28 +706,44 @@ bool V4L_camera::close()
...
+        memcpy(buffer, param.dst_image, param.dst_image_size);
+        mutex.post();
+        return true;
+    }
+    else{
+        yError()<<"usbCamera: unable to get the buffer, device unitialized";
+        mutex.post();
+        return false;
+    }
 }
```
barbalberto: Some compiler complains if there is no return at the end of a non-void function (even though all path do have a return statement). Can you try to compile it with clang?
(or refactor it to have a single return statement)

Nicogene: I checked with clang it appears this warning 
`/icub-main/src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp:1818:1: warning: control may reach end of non-void function [-Wreturn-type]
}`

PR #413, opened 2017-01-19
ActionsRenderingEngine to point far targets

pattacini:
This PR extends the way `ARE` can point at given targets that are located far from the robot. This behavior turns to be particularly useful when iCub is requested to point at human partners or objects at distance rather than objects lying on the table.

The new rpc command **`pfar`** has been added with the following syntax:
```sh
$ yarp rpc /actionsRenderingEngine/cmd:io
> pfar (x y z)
```
where `x`, `y` and `z` are the coordinates of the target.

Here's below an example of the far-pointing action:

ezgif com-optimize

cc @Tobias-Fischer @vtikha


Tobias-Fischer:
Thanks a lot @pattacini :). This is a cool feature.

PR #414, opened 2017-01-23
Python camera compatibility + exposure_absolute

barbalberto:
This PR introduce some customization for compatibility with leopard python 1300 dual camera and introduce handling of exposure_absolute control of V4L.

Since the data they provide is a custom raw format, some operation are needed in order to reconstruct a standard data type.
Basically the operation needed is a 4bit right shift and reduce the 16bit data into a 8bit. After this operation, the data is a standard 8bit bayer gr/gb ( linux code V4L2_PIX_FMT_SGRBG8).

Ruuning the usbCamera module with the option ` --camModel leopard_python` the code will apply the required fix for this camera.

Hopefully this workaround will go away when the fix will be implemented in FPGA by the producer, since doing those operation by CPU drastically reduces the framerate output.

However using the `usbCameraRaw` device, only the shift operation will be applied by the driver.
The user can leverage on the bayer carrier to do the color reconstruction on user side (on a different machine) in such a way that the computation workload will be splitted by the two machines increasing theoutput and reducing the required bandwith (since images will be sent as a bayer, using therefore 1byte per pixel instead of 3 bytes per pixel as in rgb)

` yarp connect /grabber /yarpview/img:i tcp+recv.bayer+order.gbgr`


barbalberto:
It should be merged on devel, not master.... sorry

pattacini:
@barbalberto next time you could simply edit the PR and change the base branch.

barbalberto:
Oh, really? Thanks, I didn't know

pattacini:
Yep: https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

PR #415, opened 2017-01-23
Python camera compatibility ans exposure_absolute

barbalberto:
This PR introduce some customization for compatibility with leopard python 1300 dual camera and introduce handling of exposure_absolute control of V4L.

Since the data they provide is a custom raw format, some operation are needed in order to reconstruct a standard data type.
Basically the operation needed is a 4bit right shift and reduce the 16bit data into a 8bit. After this operation, the data is a standard 8bit bayer gr/gb ( linux code V4L2_PIX_FMT_SGRBG8).

Ruuning the usbCamera module with the option --camModel leopard_python the code will apply the required fix for this camera.

Hopefully this workaround will go away when the fix will be implemented in FPGA by the producer, since doing those operation by CPU drastically reduces the framerate output.

However using the usbCameraRaw device, only the shift operation will be applied by the driver.
The user can leverage on the bayer carrier to do the color reconstruction on user side (on a different machine) in such a way that the computation workload will be splitted by the two machines increasing theoutput and reducing the required bandwith (since images will be sent as a bayer, using therefore 1byte per pixel instead of 3 bytes per pixel as in rgb)

yarp connect /grabber /yarpview/img:i tcp+recv.bayer+order.gbgr


Nicogene:
@barbalberto  I will take a look to it tomorrow because today  I'll leave after lunch :sweat_smile:

barbalberto:
@Nicogene which change? Can you point at the line(s)?

src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -768,7 +768,7 @@ bool V4L_camera::getRawBuffer(unsigned char *buffer)
     mutex.wait();
     if(configured){
         imageProcess(param.raw_image);
-        memcpy(buffer, param.dst_image, param.dst_image_size);
+        memcpy(buffer, param.tmp_image, param.dst_image_size);
```
Nicogene: @barbalberto  this change, sorry but I didn't know how to reference it :sweat_smile:

barbalberto: Well, the idea is that the `getRawBuffer` has to return the image before the color reconstruction, while the `dst_image` is the buffer containing the rgb image.

Usually the raw images are stored in `param.raw_image` but for the leopard cameras a minimum processing is still required, and the result is stored in `param.tmp_image`

So before it was wrong to use the `dst_image` because ti was the rgb image.
Now it is ok for leopard... but it is probably still wrong for standard cameras, because it shall be `param.raw_image`.
I'll check. Thanks for the catch

Nicogene: You are welcome :wink: I looked also the rest and looks good to me :+1:

Nicogene: For avoiding any kind of problem you can check: 
```c++
if(param.camModel==LEOPARD_PYTHON)
    memcpy(buffer, param.tmp_image, param.dst_image_size);
else
    memcpy(buffer, param.dst_image, param.dst_image_size);
```

PR #420, opened 2017-02-09
V4L_camera: fix potential memory leak

Nicogene:
We found and fixed a memory leak triggered when you open two devices `usbCamera`.
Please review code.


barbalberto:
This PR fixes a bug when using 2 usbCamera from the same executable

PR #421, opened 2017-02-10
Unification usbCamera and dragonFly

Nicogene:
With the new `ServerGrabber` introduced by robotology/yarp#1088, and the changes of this PR,  the streaming of RGB and RAW images are both managed by `USBCameraDriverRgb` and `DragonflyDeviceDriver2Rgb` respectively .
By default these two objects manage only the RGB images for backward compatibility.
Moreover this PR include the IRgbVisualParams in `DragonflyDeviceDriver2` and `CFWCamera_DR2_2` including also the method added by robotology/yarp#1077.

**TODO** before merging:
- [x] merge robotology/yarp#1077
- [x] merge robotology/yarp#1088
- [x] merge #415
- [x] merge #420  
- [x] rebase it removing ee2cb45 that it will be introduced already by #420 
- [x] Increase the yarp tweak number required to 7


src/libraries/icubmod/dragonfly2/common/DragonflyDeviceDriver2.cpp
```diff
@@ -40,6 +45,7 @@ int DragonflyDeviceDriver2Rgb::height () const
 	return RES(system_resources)->height();
 }
 
+#ifndef YARP_NO_DEPRECATED
```
drdanz: This is not YARP, you should not use `YARP_NO_DEPRECATED`. This is a device and not a library, so probably just removing the deprecated methods should be enough...

pattacini:
@Nicogene @barbalberto 

Hey guys, what's the impact of this PR on our daily work with iCub?
Is it 100% back-compatible and transparent or should we aware of any particular modification with respect to the previous context?

cc @vtikha

barbalberto:
It is transparent and already tested on the blacky.
Just hope it'll make things easier and cleaner for the future.

pattacini:
Perfect thanks! 👍

PR #442, opened 2017-04-26
ImageSplitter: changing the logic of the connection to the source port

Nicogene:
This PR changes the default behaviour of imageSplitter, if nothing is specified it connects to `remote` once it is launched, otherwise if `connectLater` is set it waits that the connection is made "outside" the module.

Please review code.

### DIFF
--- a/src/tools/imageSplitter/src/imageSplitter.cpp
+++ b/src/tools/imageSplitter/src/imageSplitter.cpp
@@ -26,6 +26,7 @@ ImageSplitter::~ImageSplitter()
 
 bool ImageSplitter::configure(yarp::os::ResourceFinder &rf)
 {
+    bool shouldConnect =false;
     // Check input parameters
     if(rf.check("align"))
     {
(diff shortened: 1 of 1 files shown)


src/tools/imageSplitter/src/imageSplitter.cpp
```diff
@@ -64,13 +77,16 @@ bool ImageSplitter::configure(yarp::os::ResourceFinder &rf)
...
-        yError() << "Missing 'remote' parameter. Please insert the name of the port to connect to.";
-        return false;
-    }
+        remotePortName = rf.find("remote").asString();
+        shouldConnect = true;
 
-    yarp::os::ConstString remotePortName = rf.find("remote").asString();
+    }
+    else
+        yWarning()<<"ImageSplitter: missing --remote parameter, external connection needed to " + inputPortName;
```
drdanz: This is supposed to be the normal behaviour when `--remote` is not passed, so this warninig looks a bit too much to me... Perhaps just add a

```c++
yInfo() << "ImageSplitter: waiting for connection to port" << inputPortName
```

Also this shouldn't be **here**, but after the port is actually open since it should print the **real** name of the port (i.e. the name of the port can change after `YARP_PORT_PREFIX` and `YARP_PORT_RENAME` are applied).

PR #448, opened 2017-05-19
Update xml template for using the new ServerGrabber

Nicogene:
This PR update the templates for using the new `grabberDual` through `yarpmanager`.


pattacini:
Ops, too fast? 😉

Nicogene:
The PR robotology/robots-configuration#30 for adding the new ini files has not been merged yet, but I don't think it is a big problem  because we will merge the PR and open the issue in QA on Monday :wink:

PR #450, opened 2017-05-29
usbCamera: add flip parameter for leopard_python camera

Nicogene:
This PR introduce the `-- flip` parameter for flipping the image around the *y axis*, useful for `leopard_python` cameras

TODO:
- [x] Test on `leopard_python`

Please review code.
Ready to be merged.

### DIFF
--- a/src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
+++ b/src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
@@ -339,6 +339,9 @@ bool V4L_camera::fromConfig(yarp::os::Searchable& config)
     else
         param.deviceId = config.find("d").asString();
 
+    param.flip=config.check("flip",Value("false")).asBool();
+
+
 
     if(!config.check("camModel") )
     {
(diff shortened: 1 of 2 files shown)


src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -339,6 +339,12 @@ bool V4L_camera::fromConfig(yarp::os::Searchable& config)
     else
         param.deviceId = config.find("d").asString();
 
+    if (config.find("flip").asBool())
```
pattacini: @Nicogene you should better use `check` to verify that the `flip` option is present or not, then if it does exist, you can retrieve its value.

```c++
if (config.check("flip"))
    param.flip=config.find("flip").asBool();
```

Or even better, you may force a default value:
```c++
param.flip=config.check("flip",Value(false)).asBool();
```
I don't know if `yarp::os::Value` has a proper constructor for boolean initializer; otherwise the following should make the job:
```c++
param.flip=config.check("flip",Value("false")).asBool();
```

PR #458, opened 2017-07-06
usbCamera: fix rescaling

barbalberto:
Previously the rescale did not take into account the form factor (4/3 or 16/9) of the image, so rescaling may led to stretched or compressed images.
This PR fix this error.
Also introduces the possibility to set color settings like saturation from config file or command line which was missing.

### DIFF
--- a/src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
+++ b/src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
@@ -386,7 +386,6 @@ bool V4L_camera::fromConfig(yarp::os::Searchable& config)
     param.flip=config.check("flip",Value("false")).asBool();
 
 
-
     if(!config.check("camModel") )
     {
         yInfo() << "No 'camModel' was specified, working with 'standard' uvc";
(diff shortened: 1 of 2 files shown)


barbalberto:
Tested on robot ok

src/libraries/icubmod/usbCamera/linux/V4L_camera.cpp
```diff
@@ -432,13 +476,13 @@ bool V4L_camera::fromConfig(yarp::os::Searchable& config)
...
     {
         case VOCAB_PIXEL_MONO:
-            param.pixelType = V4L2_PIX_FMT_GREY;
-            param.dst_image_size = param.width * param.height;
+            // Pixel type raw is the native one from the camera
+            param.pixelType = convertV4L_to_YARP_format(param.src_fmt.fmt.pix.pixelformat);
             break;
 
         case VOCAB_PIXEL_RGB:
+            // is variable param.pixelType really required??
```
Nicogene: This variable it is necessary if you want to use `usbCamera` for getting RGB or RAW images, setting `capabilities` in `grabberDual`.

barbalberto:
DO NOT MERGE yet, new fix is coming

barbalberto:
Done and tested on robot. Waiting for tests and then merge can be done

PR #491, opened 2018-02-01
Improve eomc

valegagge:
In this branch I did a simple refactoring of embObjMotionControl, and I didn't change any functionality.

The refactoring is not complete, but this is the first step towards a "good" implementation of embObjMotionControl.

I have already tested this branch on purple iCub


marcoaccame:
good and much needed cleaning. 

I quite liked the creation and use of a function *getBoardInfo()* to retrieve the string for the name of the board: it has removed a lot of code prone to errors all around the huge .cpp file 

On code side, only a minor comment:

1. a change is suggested for the name of the *struct eomc_encoder_t*. There is already an *enum eOmc_encoder_t* in use and this name may lead to confusion. Also, the struct has a field named *uint8_t type* which is  used in the code with values from *eOmc_encoder_t*: better using the correct type. Maybe it is worth moving this and other new types into the *namespace eomc* which you have just created to have something as:
```
namespace eomc {
    struct Encoder {
        eOmc_encoder_t type;                 /** joint encoder type*/
        double  tolerance;              /** Num of error bits passable for joint encoder */
         int     resolution;
    };
}
```


On travis side:

1. pity that there is the problem that travis checks out the master branch of yarp and compilation does not complete. It already happened with https://github.com/robotology/icub-main/pull/488. Any ideas, @barbalberto / @diegoferigo ?

diegoferigo:
@marcoaccame This is a current limitation of the strategy I implemented last time for switching to `yarp:devel` when required. As you can see, the PR build finalizes successfully, instead the PUSH build fails.

The reason is that currently `yarp:devel` is used only if either the current branch or the target branch is `devel`. In this case the current branch is `ImproveEOMC` and Travis at this stage cannot know if the target will be `master` or `devel` for this PUSH build, and it defaults to `master`.

This happens because the branch from which the PR is sourced is a branch pushed upstream. This is not the occasion to discuss the workflow of feature development, but in general my approach is to work on a branch in my personal fork. With this approach when you open a new PR from it, I think the PUSH build is not even spawned.

barbalberto:
@diegoferigo  Thanks a lot for the explanation

valegagge:
@diegoferigo Thanks for your explanation, but I need some more information to understand in detail the correct workflow. Maybe we can schedule a meeting about this.

Nevertheless, could you solve this issue provisionally?
I'm blocked on this issue.

thank you!

diegoferigo:
I can give you more details f2f, sure. At the current moment we cannot obtain a passing PUSH build, and with the current workflow on icub-main every solution would be a dirty workaround.

In my opinion, if the PR build passes, you're good to go. It represents the test against the current status of `icub-main`, and it means that your code compiles well against devel and it will succeed also after merging.

valegagge:
@marcoaccame I just commit the changes you suggested me.
In the commits I moved the parser inside the namespace eomc and not only types, so in the eomc namespace there are types, object and static function usefull to embObjMotionControl device only.

Now the code should be more readable, I hope! :)
There are some small further improvements, but I'll make them next time. (task with low prio)

@diegoferigo thanks!!!

PR #496, opened 2018-02-20
Bug fixed in checkMotionDone for all joints

valegagge:
The checkMotionDone(bool *flags) was recursive because it called itself instaed of checkMotionDoneRaw(int j, bool *flag);


barbalberto:
Merged Thanks

PR #499, opened 2018-03-12
removed variable length array which are not c++ standard.

randaz81:
removed variable length arrays which are not c++ standard.

### DIFF
--- a/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
@@ -815,7 +815,7 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
         // NOTE: (TODO) translate "fisico" and "logico" variables name in English.
 
         measureConvFactors measConvFactors_remaped (_njoints);
-        int fakeAxisMap[_njoints];
+        std::vector<int> fakeAxisMap(_njoints);
         for(int i=0; i<_njoints; i++)
         {
             measConvFactors_remaped.angleToEncoder[_axisMap[i]]  = measConvFactors.angleToEncoder[i];
@@ -838,7 +838,7 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
              trqCtrlConvFactors.init(tmpOnes.data(), tmpOnes.data());
         }
 
-        _measureConverter = new measuresConverter(_njoints,  fakeAxisMap, trqCtrlConvFactors, measConvFactors_remaped);
+        _measureConverter = new measuresConverter(_njoints,  fakeAxisMap.data(), trqCtrlConvFactors, measConvFactors_remaped);
 
 
         // 2) convert pid values from metrics units to fw units(i.e. icubDegrees)
@@ -1644,7 +1644,7 @@ bool embObjMotionControl::helper_getPosPidRaw(int j, Pid *pid)
 
 bool embObjMotionControl::helper_getPosPidsRaw(Pid *pid)
 {
-    eOmc_PID_t eoPIDList[_njoints] = {0};
+    std::vector<eOmc_PID_t> eoPIDList(_njoints);
     bool ret = askRemoteValues(eoprot_endpoint_motioncontrol, eoprot_entity_mc_joint, eoprot_tag_mc_joint_config_pidposition, eoPIDList);
     if(!ret)
     {
@@ -2181,8 +2181,7 @@ bool embObjMotionControl::checkMotionDoneRaw(int j, bool *flag)
 
 bool embObjMotionControl::checkMotionDoneRaw(bool *flag)
 {
-
-    eObool_t ismotiondoneList[_njoints] = {0};
+    std::vector <eObool_t> ismotiondoneList(_njoints);
     bool ret = askRemoteValues(eoprot_endpoint_motioncontrol, eoprot_entity_mc_joint, eoprot_tag_mc_joint_status_core_modes_ismotiondone, ismotiondoneList);
     if(false == ret)
     {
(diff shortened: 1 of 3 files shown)


valegagge:
Could we discuss f2f about this issue  in order to understand better the aim of these changes?
thanks!!

marcoaccame:
Hi everybody,
the need it just to remove code which is not C++ standard and which does not compile in window.

So, what does the code need to do? Just to store a number of values for temporary use inside an array.

It was used a variable length array which is not standard.

In this PR it is used a vector<> and the changes are good to me: they compile and work.

One objection is however that the vector<> uses heap. Yes: it uses heap to store the items, hence it calls some alloc and free inside the function. Does it reduce performance? I don't know.

But if heap is bad, i propose the following solution which compiles, works and use only the stack:


```
class embObjMotionControl ...
{
    ... 
    const unsigned int MaxNumberOfJoints = 12;
};

void myfunction()
{
    // int fakeAxisMap[_njoints]; removed because not C++

    // this variable uses stack ....
    int fakeAxisMap[embObjMotionControl::MaxNumberOfJoints];

}

```

But if we use an embObjMotionControl  with more than 12 joints, the code will crash...

randaz81:
`But if we use an embObjMotionControl with more than 12 joints, the code will crash...`
One further possibility to overcome this issue is to dynamically allocate these utility variables in the open(), use them where needed and destroy them in the close().

Please note that there is no performance overhead in **accessing** a variable length array (VLA) stored in the stack and a standard array stored in the heap with new(). In both cases, in fact, the variable is accessed through a pointer to a memory location, as mentioned in gcc documentation.

alecive:
> Allocating a std::vector each time it's gonna have a big impact in timing.

Is that true?
Or, better: is that true in a non-theoretical but realistic scenario?
Or, better: is that true in a non-theoretical but realistic scenario that has a meaningful impact on the performances?
Or, better: is that true in a non-theoretical but realistic scenario that has a meaningful impact on the performances of `icub-main`?

If that is true, is it true in the code that @randaz81 changed?

Also, what about [`std::array`](http://en.cppreference.com/w/cpp/container/array) instead? I never used it, but it seems to be in between a C-style array and an `std::vector`

 > .data() is not implemented for std::vector <bool>

Good to know!

claudiofantacci:
Hi guys, I was just reading this PR and I drop here my two cents. Hope it helps 😃 

If you want something that is not allocated in the heap, dimension is known at compile time and behave/act as a container, go for [`std:array`](http://en.cppreference.com/w/cpp/container/array) that, i quote, _combines the performance and accessibility of a C-style array with the benefits of a standard container, such as knowing its own size, supporting assignment, random access iterators, etc._

> .data() is not implemented for std::vector
> 
> Good to know!

Just a side note: `data()` [is implemented](http://en.cppreference.com/w/cpp/container/vector) in `std::vector`, but for the `std::vector<bool>` [specialization](http://en.cppreference.com/w/cpp/container/vector_bool) for which there _may_ be a space-efficient implementation.

francesco-romano:
`.data()` is a C++11 feature

barbalberto:
Just to explain the context, the whole work behind the change started because an improvement in performances was requested, when getting info for all joints from `RemoteControlBoards`.

To reach this goal different techniques was used and the biggest improvement has been achieved (and it will be, because the work is still in progress) by parallelizing the call toward the hardware.

While optimizing, we tried to see if there was a way to optimize memory usage as well, so some test have been performed.

> Or, better: is that true in a non-theoretical but realistic scenario?

Different ways of allocating temporary memory has been tested and results shown that VLA was the best option. My bad, we forgot it was not C++ compliant, so I think the most logical path is to fallback to the next best-performance option.

I think we not tested the `std::array`, so we can add it to the options and see how it behaves.

randaz81:
Guys... `std::array` cannot be used as a replacement of `std::vector` in this PR. `std::array` can be used only for arrays whose length is known at compile time. Therefore, yes, they are a valid replacement for fixed-size C arrays, but here the length is known only at run time. 

If you do not like `std::vector`, the only other viable option to me is the one that I already mentioned above: 
`One further possibility to overcome this issue is to dynamically allocate these utility variables in the open(), use them where needed and destroy them in the close().`
This because the only performance issue that I can see is the new()/delete() performed every time the function is called.

randaz81:
There is also the possibility to redefine the allocator of the `std::vector`
```
template<  class T,  class Allocator = std::allocator<T>>
 class vector;
```
This signature means that you can give to the `std::vector` your custom allocator in order to use, for example, the stack instead of the heap.
I'm not going for this solution but, the very brave ones can gofurther on , reading these QAs:
https://stackoverflow.com/questions/354442/looking-for-c-stl-like-vector-class-but-using-stack-storage
https://stackoverflow.com/questions/8049657/stack-buffer-based-stl-allocator

marcoaccame:
Hi, there is one better solution: use std::array
it is used as a vector but does not use heap, just stack.

alecive:
>    .data() is not implemented for std::vector
>
>    Good to know!
>
>Just a side note: data() is implemented in std::vector, but for the std::vector<bool> specialization for which there may be a space-efficient implementation.

@claudiofantacci : yes, what @randaz81 meant in that comment was exactly that `.data()` is not available for `std::vector<bool>`

> Guys... std::array cannot be used as a replacement of std::vector in this PR

@randaz81 what I meant (but I admittedly wrote quickly, I was just doing random things while catching a plane lol) was not to use `std::array` in place of `std::vector`. I wanted to say that you could use it in place of the `bool* isDoneList = new bool[_njoints];` that you had to use since `.data()` is not available for `std::vector<bool>`. You can then have a fully `C++` compatible code :sunglasses:

alecive:
@barbalberto good to know! As I said, I never used it, but from its description `std::array` should be both as efficient as a C-style array and C++ compliant, so I think it would be great to test its performances!

lornat75:
Question: are you sure that whether std:array is allocated in the stack or heap is not implementation specific? it would be bad to have code that behaves differently in this respect depending on the platform

randaz81:
PR tested on iCubGenova04

randaz81:
Dear all,
after f2f meeting we decided to merge this PR: it is blocking other development activities. 
I was also informed that @valegagge and @marcoaccame have an ongoing activity to improve the parallelization of multiple joints methods.
Please feel free to improve the current implementation in a future PR.

PR #513, opened 2018-04-23
Add missing icub_export_library for ethLoaderLib

Tobias-Fischer:
There is a missing `icub_export_library` in `ethLoaderLib` which causes e.g. FirmwareUpdater to fail because the library cannot be found (as it is not being installed).


Tobias-Fischer:
By the way, the missing line is copied and adapted from https://github.com/robotology/icub-main/blob/master/src/tools/canLoader/canLoaderLib/CMakeLists.txt

pattacini:
Hi @marcoaccame 
Could you look into that?

PR #520, opened 2018-05-23
Implemented CanBusMotionControl::getTorquesRaw

valegagge:
Using the controlboardWrapper with changes made in function for all joints (https://github.com/robotology/yarp/pull/1668), @Nicogene and I noticed that getTorquesRaw returns false because it was not implemented.
Therefore I implemented it like a for cicle on all joints.

@randaz81 : is there a reason why it was not implemented?


pattacini:
This fixes the huge quantity of error messages we've been observing with the robot and the simulator.
Merging for the time being; @randaz81 feel free to amend/extend it.

PR #544, opened 2018-08-10
Extend icub bindings with icub libraries and tests

towardthesea:
As title, the `PR` aims to extend the current icub bindings with more icub libraries: skinDynLib, ctrlLib, and provides some test files for `python`: test_load, test_iKin, test_ctrlLib, test_skinDynLib.

I comment out the `iDyn` libraries since nobody really use that libraries :) and I personally don't know how to test if the binding works or not :)

This should work after fixing the binding compilation with @Nicogene 's `PR` https://github.com/robotology/icub-main/pull/541

This is partly to solve the https://github.com/robotology/icub-main/issues/484 issue.

Cheers,
P

@Tobias-Fischer @nunoguedelha @traversaro @drdanz @Nicogene

### DIFF
--- a/bindings/tests/CMakeLists.txt
+++ b/bindings/tests/CMakeLists.txt
@@ -14,3 +14,6 @@ endmacro()
 file(GLOB icub_python_test_files "${CMAKE_CURRENT_SOURCE_DIR}/*.py")
 
 add_python_unit_test(test_load.py)
+add_python_unit_test(test_ctrLib.py)
+add_python_unit_test(test_iKin.py)
+add_python_unit_test(test_skinDynLib.py)
--- a/bindings/tests/CMakeLists.txt
+++ b/bindings/tests/CMakeLists.txt
@@ -1,0 +1,16 @@
+# Copyright (C) 2006-2018 Istituto Italiano di Tecnologia (IIT)
+# All rights reserved.
+#
+# This software may be modified and distributed under the terms of the
+# BSD-3-Clause license. See the accompanying LICENSE file for details.
+
+macro(add_python_unit_test pythonscript)
+  if(PYTHON_EXECUTABLE)
+    add_test(NAME ${pythonscript} COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${pythonscript})
+    set_tests_properties(${pythonscript} PROPERTIES ENVIRONMENT "PYTHONPATH=$<TARGET_FILE_DIR:${SWIG_MODULE_icub_python_REAL_NAME}>")
+  endif()
+endmacro()
+
+file(GLOB icub_python_test_files "${CMAKE_CURRENT_SOURCE_DIR}/*.py")
+
+add_python_unit_test(test_load.py)
--- a/bindings/tests/test_ctrlLib.py
+++ b/bindings/tests/test_ctrlLib.py
@@ -13,36 +13,42 @@
 # Measurement: z = [0.39,	0.50,	0.48,	0.29,	0.25,	0.32,	0.34,	0.48,	0.41,	0.45]
 # A=1, B=0, H=1, Q=0, R=0.1
 
+from __future__ import (absolute_import, division,
+                        print_function)
+from future import standard_library
+
+standard_library.install_aliases()
+
 import yarp
 
 import icub
 
 log = yarp.Log()
 
-A = yarp.Matrix(1,1)
-A[0,0] = 1.
-H = yarp.Matrix(1,1)
-H[0,0] = 1.
-Q = yarp.Matrix(1,1)
-Q[0,0] = 0.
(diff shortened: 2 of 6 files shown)


bindings/tests/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+# BSD-3-Clause license. See the accompanying LICENSE file for details.
+
+macro(add_python_unit_test pythonscript)
+  if(PYTHON_EXECUTABLE)
+    add_test(NAME ${pythonscript} COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${pythonscript})
+    set_tests_properties(${pythonscript} PROPERTIES ENVIRONMENT "PYTHONPATH=$<TARGET_FILE_DIR:${SWIG_MODULE_icub_python_REAL_NAME}>")
+  endif()
+endmacro()
+
+file(GLOB icub_python_test_files "${CMAKE_CURRENT_SOURCE_DIR}/*.py")
```
Tobias-Fischer: Is that being used?

bindings/tests/test_ctrlLib.py
```diff
@@ -0,0 +1,48 @@
...
+x = yarp.Vector(len(z))
+
+for i in range(len(z)):
+    z_vec[i] = z[i]
+    Z = yarp.Vector(1,z[i])
+    X = kal.filt(Z)
+    x[i] = X[0]
+
+log.info('Measured value                        : {:s}'.format(z_vec.toString(3,3)))
+log.debug('Corrected value with 1D Kalman filter: {:s}'.format(x.toString(3,3)))
```
Tobias-Fischer: They should be both `log.info` I suppose?

bindings/tests/test_iKin.py
```diff
@@ -0,0 +1,93 @@
...
+props.put("device", "remote_controlboard")
+props.put("local", "/" + name + "/" + arm_name)
+props.put("remote", "/" + robot + "/" + arm_name)
+
+# create remote driver
+_armDriver = yarp.PolyDriver(props)
+
+# query motor control interfaces
+_iPos = _armDriver.viewIPositionControl()
+iVel = _armDriver.viewIVelocityControl()
```
Tobias-Fischer: Any reason why this is not `_iVel`?

Tobias-Fischer: And this is not being used anyways?

bindings/tests/CMakeLists.txt
towardthesea: Actually, I only follow the same way in yarp [tests CMake](https://github.com/robotology/yarp/blob/master/bindings/python/tests/CMakeLists.txt). This requires some more changes in https://github.com/robotology/icub-main/pull/541 to really effect

bindings/tests/test_ctrlLib.py
towardthesea: I used `debug` and `info` intentionally (show off) but no problem to use `info` for both cases.

Tobias-Fischer:
Hi,
I left some inline comments. Some others:
1) I suggest to make the code compatible with both Python 2 and 3. There are not many changes required, as far as I can see it's mostly changing the `print` statement and using `future` (http://python-future.org/quickstart.html).
2) Regarding `iDyn`: If it compiles fine, I'd create the bindings. It's fine not having a test for that one.
3) Maybe separately from this PR and more appropriate for https://github.com/robotology/icub-main/pull/541 - is it worth having a Travis build configuration that compiles the bindings and runs these tests?
4) I haven't actually compiled the bindings and run those tests. I'll do it when I have some spare time.

bindings/tests/test_iKin.py
towardthesea: It has been removed :+1:

towardthesea:
Hi @Tobias-Fischer ,
1. I will add the support.
2. Regarding `iDyn`: some additional changes need to be done in order to enable it to be complied: 
- this mutual relation (between [iDyn.h](https://github.com/robotology/icub-main/blob/master/src/libraries/iDyn/include/iCub/iDyn/iDyn.h#L72)  & [iDynInv.h](https://github.com/robotology/icub-main/blob/master/src/libraries/iDyn/include/iCub/iDyn/iDynInv.h#L50) ) makes the swig compilation failed. It is surprised for me when the library can be compiled in C++. The one in `iDyn.h` needs to be removed, which I don't know how effect to `iDyn` users.
- `iDynBody.h` fails to compiled and needs deep knowledge about it to debug, which I don't really have :)

Tobias-Fischer:
Right. I suggest to open a separate issue regarding `iDyn` then, and tackle it later :+1:

towardthesea:
Changes have been tested in both Python2 & Python3

Tobias-Fischer:
Thanks - fine for me to merge after #541 will have been merged.

PR #545, opened 2018-08-10
Try to fix OS X build

Tobias-Fischer:
Following https://github.com/robotology/icub-hri/commit/68bc4ad1d298ed975b6c0c7e6fc1843cd1e74eb2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ before_script:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update; fi
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get --assume-yes --force-yes install cmake libeigen3-dev icub-common libv4lconvert0 libv4l-dev; fi
   # macOS dependencies
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update; fi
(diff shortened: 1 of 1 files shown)


Tobias-Fischer:
I think this can be merged. It is using an older OS X image than the default one (xcode8.3 instead of the default xcode9.4 [see https://docs.travis-ci.com/user/reference/osx/#Xcode-9.4]), but I guess it's better than failing.

francesco-romano:
Not sure the correct way is to downgrade the Xcode version.
What is the real error on travis?
Is this one?
```
rm: /usr/local/bin/f2py: No such file or directory
```
(which btw I might have added one year ago)

pattacini:
FYI there also exists an unmerged PR along the same line #474.
Would be great if you can consider those modifications as well and integrate them if needed; otherwise, please purge that stale branch.

Tobias-Fischer:
@francesco-romano: You are right, without those two lines it works fine using the newer XCode image.
@pattacini: #474 also tackles another problem - using the dependency list of the icub package rather than specifying and installing the dependencies manually. I am not sure whether that works - I can give it a go later.

Tobias-Fischer:
So using the `icub-main` formula dependencies doesn't work out of the box, and I suggest to fix this separately (or if someone wants, please feel free to take over this PR).
The last commit should work, and it would be great if the PR gets merged (best to squash the commits while merging).

PR #547, opened 2018-08-22
Wrap optimization library in bindings

Tobias-Fischer:
Following the discussion in https://github.com/robotology/icub-main/issues/484 this wraps most of the optimization library. To make it usable, we need to apply a typemaps for methods that "return" values by reference (not sure whether this also applies to other libraries). I added one example for the calibrate() method (and a test for it). I won't have time (and patience) to wrap all the other methods which pass by reference.

Some notes:
1) Swig 3.0.12 is required (3.0.8 throws some errors when trying to access the data in Python) for the typemap stuff. A check should be added.
2) The typemap currently only checks the parameter types + names, but not the method name. I can't seem to get it to work if I add the method name.
3) There might be a memory leak in the typemap; I'm not sure whether the pointers are ever freed. This thread might give some insight: https://stackoverflow.com/questions/7223327/what-does-the-last-argument-to-swig-newpointerobj-mean
4) The typemap only works for python (no check in place currently), I'm not sure how to wrap it for other languages.

Maybe @towardthesea can take over from here ;-)?

### DIFF
--- a/bindings/icub.i
+++ b/bindings/icub.i
@@ -48,6 +48,12 @@ using namespace yarp::sig;
 #include <iCub/skinDynLib/skinContact.h>
 #include <iCub/skinDynLib/skinContactList.h>
 
+// optimization
+#include <iCub/optimization/affinity.h>
+#include <iCub/optimization/algorithms.h>
+#include <iCub/optimization/calibReference.h>
+#include <iCub/optimization/matrixTransformation.h>
+#include <iCub/optimization/neuralNetworks.h>
 
 // iDyn
 //#include <iCub/iDyn/iDynInv.h>
@@ -91,6 +97,14 @@ using namespace yarp::sig;
 %include <iCub/skinDynLib/rpcSkinManager.h>
 %include <iCub/skinDynLib/skinContact.h>
 %include <iCub/skinDynLib/skinContactList.h>
+
+
+// optimization
+%include <iCub/optimization/matrixTransformation.h>
+%include <iCub/optimization/affinity.h>
+%include <iCub/optimization/algorithms.h>
+%include <iCub/optimization/calibReference.h>
+%include <iCub/optimization/neuralNetworks.h>
 
 
 // iDyn
--- a/bindings/icub.i
+++ b/bindings/icub.i
@@ -6,6 +6,8 @@
 
 %include "std_string.i"
 %include "std_vector.i"
+%include "std_deque.i"
+%include "typemaps.i"
 
 %import "yarp.i"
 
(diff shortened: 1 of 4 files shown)


towardthesea:
Hi @Tobias-Fischer ,
Thanks for the contributions. I have some following comments:
- Maybe you should also include the directory of `optimization` libraries like `include_directories(${CMAKE_SOURCE_DIR}/src/libraries/optimization/include)`. I got error during compilation regarding to finding the related files.
- I got error with `import icub` though my `.bashrc` includes `PYTHONPATH` pointing to the built `lib` folder (where `_icub.so` is). This wasn't error before. 
- I got error with these lines:
```
    $result= SWIG_Python_AppendOutput ($result, SWIG_NewPointerObj(t_out1, SWIGTYPE_p_yarp__sig__Matrix, 0 |  0 ));
    $result= SWIG_Python_AppendOutput ($result, SWIG_NewPointerObj(t_out2, SWIGTYPE_p_yarp__sig__VectorOfT_double_t, 0 |  0 ));
    $result= SWIG_Python_AppendOutput ($result, PyFloat_FromDouble(*$3));
``` 
when compiling with other languages (beside python). Maybe @Nicogene can have some suggestions?

Tobias-Fischer:
Hi @towardthesea,
- Re the PYTHONPATH: You should not point to the `lib` directory anymore. There is now a `dist-packages/python2.7` directory somewhere which is the one to use.
- Re the `SWIG_Python_AppendOutput`: Yes, there should be a guard that this only happens in the Python bindings. Actually, seeing https://github.com/swig/swig/search?utf8=✓&q=SWIG_AppendOutput&type= it seems that the more generic `SWIG_AppendOutput` is available for Ruby as well. Feel free to change accordingly :)

Best, Tobi

towardthesea:
Hi @Tobias-Fischer 
> Re the PYTHONPATH: You should not point to the lib directory anymore. There is now a dist-packages/python2.7 directory somewhere which is the one to use.

I guess you mean for the separated compilation of `bindings` rather than together with `icub-main`. For separated way, when `make install` is possible, your solution works like charm :+1: .

Tobias-Fischer:
Hi,
1. I never compile the bindings separately (I think this has been discouraged for a while). 
2. Also, the bindings are also installed when you do `make install` within `icub-main/build`.
3. Please try and point your PYTHONPATH to `icub-main/build`. It should find the `_icub.so` thanks to setting the RPATH.
Best, Tobi

bindings/icub.i
```diff
@@ -93,6 +101,33 @@ using namespace yarp::sig;
...
 
+%typemap(in, numinputs=0) (yarp::sig::Matrix &H, yarp::sig::Vector &s, double &error) {
+    yarp::sig::Matrix t1;
+    yarp::sig::Vector t2;
+    double t3;
+    $1= &t1;
+    $2= &t2;
+    $3= &t3;
+}
+
```
Nicogene: I'd add something like 
```
#ifdef SWIG_PYTHON
..... 
#endif
```
I am afraid that swig will not like this "python" directive while trying to compile bindings for other languages

towardthesea: :+1:  Thanks @Nicogene . Changes are pushed

Nicogene:
Merged, thanks!

Tobias-Fischer:
Hmmm .. are you guys sure it was ready to be merged? At least a check for the SWIG version should be added (see my first note in the original post), otherwise things might go wrong for some people ..

Nicogene:
Unfortunately we have not a way for test them with different version of swig, actually there are not unit test in `icub-main` as far I know. :sweat_smile: 
If someone with other version of swig have problems, we will fix it, I'm afraid it is the only way.

PR #556, opened 2018-10-05
Fix for iCub v3 torso kinematics

pattacini:
This PR addresses the need for handling the iCub v3 torso kinematics correctly.

@ale-git please give it a look.


pattacini:
Good then, thanks @ale-git 👍 
I would modify `demoRedBall` accordingly too.

PR #567, opened 2019-01-15
BUGFIX: FirmwareUpdater now correctly manages the tx of calibrationDa…

marcoaccame:
…taSNxxx.dat file also in cfw2 drivers.

the problem: if we attempt to program the calibration data of the strain in the case of cfw2 driver,
the can frames are sent as long bursts and the boards loses some of them because of its short rx buffer.

the solution: added small delays to split the long bursts.

we need to do the bugfix in master to allow a cfw2-based robot with master to re-program a strain board.


pattacini:
@marcoaccame you can merge it at your convenience

marcoaccame:
thanks

PR #572, opened 2019-03-29
Fix camCalibs after the introduction of move semantic in YARP images

Nicogene:
It fixes https://github.com/robotology/yarp/issues/1979 .

The first `toCvMat(out)` has been removed since it is not necessary, and it makes the yarp and cv output image memory shared.
I am not sure but it seems that `cv::remap` does some memory changes that made crash the program while trying to write the yarp image.
This crash does not happen without using `cv::remap`.

I tested it on my laptop with `usbCamera` and now it works. 
 
Please review code.


Nicogene:
This db6a565 should be cherry-picked also in `master`

Nicogene:
I need to run more tests before merging

Nicogene:
Successfully tested on `iCubErzelli02`

PR #574, opened 2019-04-10
Remove usbCamera

drdanz:
Moved to YARP, see robotology/yarp#1995


Nicogene:
Do we should tweak the yarp version and then increment the yarp version required here in iCub?
 In this way if someone update `icub-main` and don't find the `usbCamera` has to first update `yarp`, and ta-da it is there.

PR #580, opened 2019-06-03
EmboObjMotionControl: fix checkMotionControl(bool *flag)

valegagge:
This fix is concerning yarp issue [#2027](https://github.com/robotology/yarp/issues/2027) and needs yarp PR [#2035](https://github.com/robotology/yarp/pull/2035) to work properly


drdanz:
Merged, thanks.

PR #594, opened 2019-08-12
Introduced IMU 2.6

pattacini:
As per request #570, this PR introduces IMU version **`2.6`**.
Changes are regarded with:
- `iKin` to declare the new kinematics.
  Differently from what I planned, HN matrices are accumulated to adapt to the current code.
- `iKinGazeCtrl` to extend the set of allowed IMU's.

### DIFF
--- a/src/libraries/iKin/src/iKinFwd.cpp
+++ b/src/libraries/iKin/src/iKinFwd.cpp
@@ -2430,7 +2430,7 @@ void iCubEye::allocate(const string &_type)
         pushLink(new iKinLink(    0.0,  -0.034, -M_PI/2.0,       0.0, -35.0*CTRL_DEG2RAD, 15.0*CTRL_DEG2RAD));
         pushLink(new iKinLink(    0.0,     0.0,  M_PI/2.0, -M_PI/2.0, -50.0*CTRL_DEG2RAD, 50.0*CTRL_DEG2RAD));
     }
-    else if ((getType()=="right_v2") || (getType()=="right_v2.5"))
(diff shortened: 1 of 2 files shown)


Nicogene:
I tried it on `iCubGenova02` using `iKinGazeCtrl` . Looking left and right the neck moves ok, the problem are the eyes that start do crazy movements.

Any idea what could be the problem?
It seems a sign problem somewhere :sweat_smile:

src/libraries/iKin/src/iKinFwd.cpp
```diff
@@ -2674,6 +2674,19 @@ void iCubInertialSensor::allocate(const string &_type)
         setHN(getHN()*HN);
     }
 
+    if (getType()=="v2.6")
+    {
+        Matrix HN=zeros(4,4);
+        HN(0,3)=0.323779;
+        HN(1,3)=-0.139537;
+        HN(2,3)=0.72;
```
Nicogene: Should be divided by 10 more ?

pattacini: Feel free to fix this 👍

pattacini: Feel free to fix this 👍

pattacini:
@Nicogene give the matrix HN a second look: I might have reported it wrongly.

Nicogene:
Ok I will modify and test it

Nicogene:
Probably also other part of the library ([example](https://github.com/robotology/icub-main/blob/master/src/libraries/iKin/src/iKinFwd.cpp#L2433)) should manage this new `head_version`

PR #598, opened 2019-09-03
Fix memory usage in ParametricCalibratorEth

valegagge:
In this PR I fix a wrong use of memory allocation. 
Every time the calibrate function is called, two new heap allocations are done. (see [here](https://github.com/robotology/icub-main/blob/55154efbb1d94983598f85e270c6ad145a3ba4de/src/libraries/icubmod/parametricCalibratorEth/parametricCalibratorEth.cpp#L588) )

I moved such allocations in the `open` function, so they are performed one  time only. The delete is done in `close` function


randaz81:
Nice catch! I consider this a bugfix, so you are allowed to merge your PR into _master_, if you want to.

randaz81:
rebase in progress

PR #599, opened 2019-09-09
Added option to calibrate standalone stereo setup

Iaxama:
Hi,

I am making this pull request because recently we had to calibrate a standalone stereo camera setup. The code changes include a new flag in the stereoCalib module called "standalone" which, if set, skips the parts of the module that assume the icub to be up and running. Additionally, I allow the two cameras to have different resolution when computing the extrinsic parameters, as long as the intrinsic ones are computed beforehand.

### DIFF
--- a/src/tools/stereoCalib/src/stereoCalibThread.cpp
+++ b/src/tools/stereoCalib/src/stereoCalibThread.cpp
@@ -33,8 +33,7 @@ stereoCalibThread::stereoCalibThread(ResourceFinder &rf, Port* commPort, const c
     int tmp=stereoCalibOpts.check("MonoCalib", Value(0)).asInt();
     this->stereo= tmp?false:true;
     this->camCalibFile=rf.getHomeContextPath().c_str();
-    this->standalone = rf.check("standalone") &&
-                  rf.check("standalone", yarp::os::Value(true)).asBool();
+    this->standalone = rf.check("standalone");
     string fileName= "outputCalib.ini"; //rf.find("from").asString().c_str();
 
     this->camCalibFile=this->camCalibFile+"/"+fileName.c_str();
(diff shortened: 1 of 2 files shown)


src/tools/stereoCalib/src/stereoCalibThread.cpp
```diff
@@ -33,7 +33,8 @@ stereoCalibThread::stereoCalibThread(ResourceFinder &rf, Port* commPort, const c
     int tmp=stereoCalibOpts.check("MonoCalib", Value(0)).asInt();
     this->stereo= tmp?false:true;
     this->camCalibFile=rf.getHomeContextPath().c_str();
-
+    this->standalone = rf.check("standalone") &&
```
pattacini: What's that? The check is replicated twice: one should be sufficient 😉

src/tools/stereoCalib/src/stereoCalibThread.cpp
```diff
@@ -43,6 +44,30 @@ stereoCalibThread::stereoCalibThread(ResourceFinder &rf, Port* commPort, const c
 
 bool stereoCalibThread::threadInit()
 {
+/* FOR DEBUG LOAD IMAGES FROM FILE
```
pattacini: Please, get rid of all these comments.

src/tools/stereoCalib/src/stereoCalibThread.cpp
```diff
@@ -219,7 +244,7 @@ void stereoCalibThread::stereoCalibRun()
                     stereoCalibration(imageListLR, this->boardWidth,this->boardHeight,this->squareSize);
 
                     yInfo(" Saving Calibration Results... \n");
-                    updateIntrinsics(Left.cols,Left.rows,Kright.at<double>(0,0),Kright.at<double>(1,1),Kright.at<double>(0,2),Kright.at<double>(1,2),DistR.at<double>(0,0),DistR.at<double>(0,1),DistR.at<double>(0,2),DistR.at<double>(0,3),"CAMERA_CALIBRATION_RIGHT");
+                    updateIntrinsics(Right.cols,Right.rows,Kright.at<double>(0,0),Kright.at<double>(1,1),Kright.at<double>(0,2),Kright.at<double>(1,2),DistR.at<double>(0,0),DistR.at<double>(0,1),DistR.at<double>(0,2),DistR.at<double>(0,3),"CAMERA_CALIBRATION_RIGHT");
```
pattacini: Woh.. changing from `left` to `right` seems a big modification. From the PR review dashboard, it's almost impossible to judge it. Could you add here some hints on that?

pattacini:
@Iaxama I've also changed the PR to be merged against `devel` instead of `master`.

src/tools/stereoCalib/src/stereoCalibThread.cpp
Iaxama: That's just to use the parameter as a flag and write in command line _--standalone_ rather than _--standalone true/false_

src/tools/stereoCalib/src/stereoCalibThread.cpp
Iaxama: Yes, sorry about that.

src/tools/stereoCalib/src/stereoCalibThread.cpp
Iaxama: That function call is actually referred to the intrinsic parameter update for the Right camera. It did not make the difference so far, because the two images have also been assumed to have the same size (cols and rows). However in this pull request we are not making this assumption anymore and allowing the two images to have different sizes, as long as the intrinsic parameters are calibrated separately beforehand (throwing an error otherwise).

src/tools/stereoCalib/src/stereoCalibThread.cpp
pattacini: Interesting; did you check it's behaving as expected?

src/tools/stereoCalib/src/stereoCalibThread.cpp
pattacini: Still need to push?

src/tools/stereoCalib/src/stereoCalibThread.cpp
pattacini: Ok then! It's been actually a bug fix too 👍

src/tools/stereoCalib/src/stereoCalibThread.cpp
Iaxama: Pleas check the latest commit a2692d30bf9e5da41122c95dc45bbdd13e978713

PR #605, opened 2019-09-26
Port the imu of iCub_SIM to the MAS interfaces`

Nicogene:
Moreover it instantiates the `multipleanalogsensorsserver` that wraps the new icub sim device that expose the imu.

The port `/icub/inertial` has been maintained for backward compatibilty.

I only checked the number coming from `measures:o`, they seems ok, to be tested with `iCubGui`.


Nicogene:
Tested successfully with `iCubGui` :+1:

pattacini:
It's mergeable for me then 👍

PR #614, opened 2019-10-28
EmbObj uses std::mutex + std::condition_variable

pattacini:
This PR comes along #612 and #613 with the aim to replace the following components:
- `yarp::os::Semaphore`
- `yarp::os::Mutex`
- `yarp::os::LockGuard`
- `yarp::os::Event`
with their STD counterparts:
- `std::mutex`
- `std::condition_variable`
- `std::lock_guard`

The PR applies the required changes only to **`embObj`**. 

#### to wait:
```c++
std::mutex mtx_event;

cv_event.wait(lck);
```
#### to post:
```c++
cv_event.notify_one();
```

There exists also the equivalent syntax that lets wait until a timeout. See for ref. https://github.com/robotology/assistive-rehab/blob/master/modules/navController/src/main.cpp#L459-L465.

### DIFF
--- a/src/libraries/icubmod/embObjAnalog/embObjAnalogSensor.cpp
+++ b/src/libraries/icubmod/embObjAnalog/embObjAnalogSensor.cpp
@@ -1182,11 +1182,10 @@ int embObjAnalogSensor::read(yarp::sig::Vector &out)
 {
     // This method gives analogdata to the analogServer
 
-      mutex.wait();
+      std::lock_guard<std::mutex> lck(mtx);
 
       if (!analogdata)
       {
-          mutex.post();
           return false;
       }
 
@@ -1212,7 +1211,6 @@ int embObjAnalogSensor::read(yarp::sig::Vector &out)
                   counterError++;
               } break;
           }
-          mutex.post();
           return status;
       }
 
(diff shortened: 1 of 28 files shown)


marcoaccame:
Just a memo.

I have tested on iCubGenova02 but yarprobotinterface gives me runtime errors the likes of:

```
||| finding paths [plugins]
error while opening /home/acemor/robotology/icub-main/build/lib/iCub/embObjStrain.so:
  /home/acemor/robotology/icub-main/build/lib/libethResources.so.0: undefined symbol:
 _ZN3eth13TheEthManager5txSemE
Cannot load plugin from shared library (embObjStrain)
(embObjStrain: cannot open shared object file: No such file or directory)
yarpdev: ***ERROR*** could not find device <embObjStrain>
```

The same happens on my laptop.

I will analize the problem in the following days.

marcoaccame:
Hi @triccyx ...
following your suggestion, I could not help but looking at the code once more.
Yes, there is a variable which is a static member of a class with no explicit initialization in cpp code. 
Thanks.

pattacini:
@marcoaccame I've rebased this branch on top of the latest `devel`.

pattacini:
@marcoaccame yet another rebase on top of `devel`.

marcoaccame:
the commit https://github.com/robotology/icub-main/pull/614/commits/df59bf4b3fabb771e5f8a0723a8e999e2ca19ca9 has solved the problem.
yarprobotinterface runs on my laptop (w/ robots-configuration/experimentalSetups/singleETHboard/icub_only_rfe.xml) and on iCubGenova02

It can be merged

PR #615, opened 2019-10-28
iCub_SIM uses std::mutex + std::condition_variable

pattacini:
This PR comes along #612, #613 and #614 with the aim to replace the following components:
- `yarp::os::Semaphore`
- `yarp::os::Mutex`
- `yarp::os::LockGuard`
- `yarp::os::Event`
with their STD counterparts:
- `std::mutex`
- `std::condition_variable`
- `std::lock_guard`

The PR applies the required changes only to **`iCub_SIM`**. 

#### to wait:
```c++
std::mutex mtx_event;

cv_event.wait(lck);
```
#### to post:
```c++
cv_event.notify_all();
```

There exists also the equivalent syntax that lets wait until a timeout. See for ref. https://github.com/robotology/assistive-rehab/blob/master/modules/navController/src/main.cpp#L488-L494.


pattacini:
`iCub_SIM` doesn't start up correctly. I'll have a look into it asap 👀 🐞

pattacini:
🚦 green light 👍 
Although it might be worth testing this on Linux too.

pattacini:
Tested on Linux too ✔️

pattacini:
Actually, 2cd1e7128f0b1aff7f0582a2e9e84702a7b6c946 was missing.

PR #618, opened 2019-10-31
icubmod uses std::mutex + std::condition_variable (4)

pattacini:
This PR comes along #612 #613 #614 #615 #616 #617 with the aim to replace the following components:
- `yarp::os::Semaphore`
- `yarp::os::Mutex`
- `yarp::os::LockGuard`
- `yarp::os::Event`
with their STD counterparts:
- `std::mutex`
- `std::condition_variable`
- `std::lock_guard`

The PR applies the required changes to a subset of `icubmod` components.

#### to wait:
```c++
std::mutex mtx_event;

cv_event.wait(lck);
```
#### to post:
```c++
cv_event.notify_one();
```

There exists also the equivalent syntax that lets wait until a timeout. See for ref. https://github.com/robotology/assistive-rehab/blob/master/modules/navController/src/main.cpp#L459-L465.


pattacini:
Tested completed successfully.

PR #622, opened 2019-11-12
Add the possibility to change the color of the face expression

Nicogene:
It has been added a new rpc request:
```bash
$ set col <color>
```
Where color is:

One of the following: black, white, red, lime, blue, yellow, cyan, magenta, silver, gray, maroon, olive, green, purple, teal, navy.

:warning: 
This new feature work only if the icub mount the new rfe board.

Please review code.

### DIFF
--- a/app/faceExpressions/conf/emotions_rfe.ini
+++ b/app/faceExpressions/conf/emotions_rfe.ini
@@ -16,3 +16,22 @@ E8   evi 05h 05h 05h 70h
 E9   shy 06h 06h 06h 70h
 E10  cun 07h 07h 07h 38h
 
+//number of colors
+colors 16
+
+C0   black   00h
+C1   white   01h
+C2   red     02h
+C3   lime    03h
+C4   blue    04h
+C5   yellow  05h
+C6   cyan    06h
+C7   magenta 07h
+C8   silver  08h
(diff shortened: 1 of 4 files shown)


Nicogene:
Tested successfully on `iCubGenova02
`

Nicogene:
We added also add the possibility to set the brightness of the leds
``` bash
$ set brig <val>
```
Where val goes from 0 to 5.

Nicogene:
We added the possibility to specify bitmasks for defining the expression

E.g.:
``` bash
$ set mask (<color> <mask_name> <bright>) (<color> <mask_name> <bright>) (<color> <mask_name> <bright>)
```
It defines respectively the bitmask of the left eyebrow, right eyebrow and mouth.
The bitmask <mask_name> can be found/defined in the /faceExpressions/emotions_rfe.ini.

E.g.: 

BM_M1  (0 2 6 9 10 12 14 18)           //00045645h // happy

This is the bitmask for the happy face, it is defined by a list of integer that specifies which led are on.

Nicogene:
@marcoaccame and me tested all the new and old functionalities and it works.
Merging then.

BTW for exploiting this new features it is needed to have a rfe board and to update the firmware to the latest version.

PR #635, opened 2020-01-16
Fix iCubGui red ball, it is now actually a ball

Nicogene:
It adresses https://github.com/icub-tech-iit/task-force-miscellanea/issues/25

image


drdanz:
I suspect that the ball might not be positioned in the centre of the cube, but in the corner instead.

Also this is a workaround, the rpc call should probably have an option to make the object spherical instead of a box

PR #642, opened 2020-02-20
[WIP] Cleanup: remove trailing spaces.

Nicogene:
Command used: `for f in $(git ls-files); do sed -i 's/[ \t]*$//' "$f"; done`


Nicogene:
Actually it changed also png files and I am afraid that it corrupted

pattacini:
Not really in favor of this, sorry...

pattacini:
This PR proposes modifications that are environment-dependent and there's no guarantee that we won't generate in the future further changes not in accordance with this policy.

I'm going to close this.

PR #647, opened 2020-03-10
Bindings target now Python-3

pattacini:
This PR cleans up bindings in order to use Python-3.

If Python-3 is not found, an attempt is made to rely on Python-2 instead.

Thanks @diegoferigo for the [heads-up][1].
A test to verify the PR has been done successfully.

Be careful as https://github.com/robotology/icub-main/issues/645 has not been solved yet.

[1]: https://github.com/robotology/QA/issues/391#issuecomment-597006107

### DIFF
--- a/bindings/CMakeLists.txt
+++ b/bindings/CMakeLists.txt
@@ -53,12 +53,10 @@ if(CREATE_PYTHON)
     find_package(Python REQUIRED COMPONENTS Interpreter Development)
     set(CMAKE_SWIG_FLAGS "-Wall;-module;icub;-threads")
 
-    include_directories(SYSTEM ${Python_INCLUDE_DIRS})
     swig_add_library(icub_python
                      LANGUAGE python
                      SOURCES icub.i)
-    target_include_directories(${SWIG_MODULE_icub_python_REAL_NAME} SYSTEM PRIVATE ${Python_INCLUDE_DIRS})
-    target_link_libraries(${SWIG_MODULE_icub_python_REAL_NAME} ${Python_LIBRARIES})
+    target_link_libraries(${SWIG_MODULE_icub_python_REAL_NAME} Python::Python)
     set_target_properties(${SWIG_MODULE_icub_python_REAL_NAME} PROPERTIES OUTPUT_NAME "_icub")
 
     # installation path is determined reliably on most platforms using distutils
--- a/bindings/CMakeLists.txt
+++ b/bindings/CMakeLists.txt
@@ -2,12 +2,12 @@
 # Authors: Paul Fitzpatrick
 # CopyPolicy: Released under the terms of the GNU GPL v2.0.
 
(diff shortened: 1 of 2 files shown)


bindings/CMakeLists.txt
```diff
@@ -50,29 +50,21 @@ endif("${SWIG_VERSION}" VERSION_LESS "${MIN_SWIG_VERSION}")
...
-
-    include_directories(SYSTEM ${PYTHON_INCLUDE_DIRS})
+    include_directories(SYSTEM ${Python_INCLUDE_DIRS})
     swig_add_library(icub_python
                      LANGUAGE python
                      SOURCES icub.i)
-    target_include_directories(${SWIG_MODULE_icub_python_REAL_NAME} SYSTEM PRIVATE ${PYTHON_INCLUDE_DIRS})
-    target_link_libraries(${SWIG_MODULE_icub_python_REAL_NAME} ${PYTHON_LIBRARIES})
+    target_include_directories(${SWIG_MODULE_icub_python_REAL_NAME} SYSTEM PRIVATE ${Python_INCLUDE_DIRS})
+    target_link_libraries(${SWIG_MODULE_icub_python_REAL_NAME} ${Python_LIBRARIES})
```
diegoferigo: If it works, the following would be more compliant to Modern CMake:

```suggestion
    target_link_libraries(${SWIG_MODULE_icub_python_REAL_NAME} Python::Python)
```

With this modification you can probably also remove the `include_directory` of line 56.

bindings/CMakeLists.txt
```diff
@@ -50,29 +50,21 @@ endif("${SWIG_VERSION}" VERSION_LESS "${MIN_SWIG_VERSION}")
...
     set_target_properties(${SWIG_MODULE_icub_python_REAL_NAME} PROPERTIES OUTPUT_NAME "_icub")
 
     # installation path is determined reliably on most platforms using distutils
-    execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))"
+    execute_process(COMMAND ${Python_EXECUTABLE} -c "from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))"
                     OUTPUT_VARIABLE PYTHON_INSTDIR
-                    OUTPUT_STRIP_TRAILING_WHITESPACE )
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)
```
diegoferigo: For the time being this is ok, however let's keep in mind that the new FindPython provides something similar. Though, there are few caveats as discussed in https://github.com/robotology/robotology-superbuild/pull/359#issuecomment-592956444.

pattacini: Thanks for pointing this out, although I would rather stay more conservative here 👍 
Feel free to contribute in this sense anyway.

PR #650, opened 2020-04-02
Fix/cleanup of dependencies of embobj devices and tools

Nicogene:
This PR add `target_include_directories` to `iCubDev` and it does some cleanup in the `CmakeLists.txt` in order that each target link only the libraries it need.

It should fix the regression on Ubuntu builds of #649 (see https://github.com/robotology/robotology-superbuild/issues/372#issue-592419756)

I tested it on my machine and it compiles fine from a clean setup


Nicogene:
All the `icub_fimware_shared` stuff are not in the CI, so no need to wait that builds finish.

PR #668, opened 2020-06-26
emotioninterface: add the command for controlling the eyelids for the rfe configuration

Nicogene:
This PR add the commands for controlling the eylids in the rfe iCub.

In particular in `emotion_rfe.ini` the eyelids column has been changed, the value is no more HEX but a representation of `% open`
E.g.

`00p` -> 0% open
`01p` -> 10% open
`05p` -> 50% open
`10p` -> 100% open

It has to be tested on both old and new robots.

### DIFF
--- a/app/faceExpressions/conf/emotions_rfe.ini
+++ b/app/faceExpressions/conf/emotions_rfe.ini
@@ -2,19 +2,19 @@
 emotions 10
 
 //configuration of facial expressions
-//index name Left-Eyebrow(HEX) Right-Eyebrow(HEX) Mouth(HEX) Eyelids(% open)
+//index name Left-Eyebrow(HEX) Right-Eyebrow(HEX) Mouth(HEX) Eyelids(% open, e.g 00=0%, 05=50%, 10 = 100%)
 //index starts at 1
 
-E1   neu 00h 00h 00h 38h
-E2   ta1 **h **h 00h **h
-E3   ta2 **h **h 03h **h
-E4   hap 01h 01h 01h 70h
-E5   sad 02h 02h 02h 38h
-E6   sur 03h 03h 03h 38h
-E7   ang 04h 04h 04h 70h
-E8   evi 05h 05h 05h 70h
-E9   shy 06h 06h 06h 70h
-E10  cun 07h 07h 07h 38h
+E1   neu 00h 00h 00h 00p
+E2   ta1 **h **h 00h **p
+E3   ta2 **h **h 03h **p
+E4   hap 01h 01h 01h 10p
+E5   sad 02h 02h 02h 00p
+E6   sur 03h 03h 03h 00p
+E7   ang 04h 04h 04h 10p
+E8   evi 05h 05h 05h 10p
+E9   shy 06h 06h 06h 10p
+E10  cun 07h 07h 07h 00p
 
 //number of colors
 colors 16
--- a/src/core/emotionInterface/emotionInterfaceModule.cpp
+++ b/src/core/emotionInterface/emotionInterfaceModule.cpp
@@ -551,13 +551,7 @@ bool EmotionInterfaceModule::setEyelids(const std::string cmd)
         return true;  //leave it in the same state
     if (_iPos)
     {
-        std::string percentage_str{ _emotion_table[i].eli[0] };
-        percentage_str += '.';
-        percentage_str += _emotion_table[i].eli[1];
-        auto percentage = std::stod(percentage_str);
-        if(percentage > 1.0)
-            percentage = 1.0;
-        auto target_pos = (1.0 - percentage) * (_max-_min); // because min-> open, max->closed
+        auto target_pos = getEyelidsTarget(_emotion_table[i].eli[0], _emotion_table[i].eli[1]);
         res = _iPos->positionMove(_joint_eylids, target_pos);
     }
     else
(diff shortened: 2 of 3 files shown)


app/faceExpressions/conf/emotions_rfe.ini
```diff
@@ -2,19 +2,19 @@
...
+E1   neu 00h 00h 00h 00p
+E2   ta1 **h **h 00h **p
+E3   ta2 **h **h 03h **p
+E4   hap 01h 01h 01h 10p
+E5   sad 02h 02h 02h 00p
+E6   sur 03h 03h 03h 00p
+E7   ang 04h 04h 04h 10p
+E8   evi 05h 05h 05h 10p
+E9   shy 06h 06h 06h 10p
+E10  cun 07h 07h 07h 00p
```
pattacini, line 17: Wouldn't it be better to use 3 digits to represent percentages so that we could have 100 --> 100%?
Is there any limitation in the number of digits?

src/core/emotionInterface/emotionInterfaceModule.cpp
```diff
@@ -500,19 +537,29 @@ bool EmotionInterfaceModule::setMouth(const std::string cmd)
...
+        auto target_pos = std::stod(percentage_str) * (_max-_min);
+        res = _iPos->positionMove(_joint_eylids, target_pos);
+    }
+    else
+    {
+        cmdbuffer[0] = 'S';
+        cmdbuffer[1] = _emotion_table[i].eli[0];
+        cmdbuffer[2] = _emotion_table[i].eli[1];
+        res = writePort(cmdbuffer);
+    }
```
pattacini, line 560: If I'm not mistaken, this `else` should check if the robot doesn't have the RFE (in case we'll raise only a warning for RFE robots w/ motors off).

src/core/emotionInterface/emotionInterfaceModule.cpp
```diff
@@ -243,6 +245,35 @@ bool EmotionInterfaceModule::configure(ResourceFinder& config){
...
+        ok &= _poly.view(iCtrlLim);
+        if (iCM)
+            ok &= iCM->setControlMode(_joint_eylids, VOCAB_CM_POSITION);
+        if (iCtrlLim)
+            ok &= iCtrlLim->getLimits(_joint_eylids, &_min, &_max);
+        if (!ok)
+        {
+            yError() << "Fail to configure correctly the remote_controlboard";
+            return false;
+        }
```
pattacini, line 272: Didn't we say that for RFE robot w/ motors off we should only raise a warning? This should cover the not infrequent conditions when at the test bench only LED's will need verification.

cc @marcoaccame

src/core/emotionInterface/emotionInterfaceModule.cpp
Nicogene: In this `else` fall:

- rfe robots that have issues in the face motors(broken or turn off)
- old robots that commands the eyelids through serial.

Unfortunately there is not a clear way to understand inside the module if the robot has or not the rfe

src/core/emotionInterface/emotionInterfaceModule.cpp
Nicogene: If I did it correctly it should try to open the polidriver, and the fail of this open is **not** fatal.
It is **fatal** if once open it can't be configured correctly(view of the interfaces), that is very rare

pattacini: Ok then! I would do a specific test on this condition.

src/core/emotionInterface/emotionInterfaceModule.cpp
pattacini: Ok, if we don't have an easy way to check whether a robot is RFE, then we stick to this 👍

app/faceExpressions/conf/emotions_rfe.ini
Nicogene: The third character it is ignored in the parsing phase of the ini file, for using it I have handle in a different way the configuration phase for the old robot and the new one.
In this way the parsing part is not touched, just the execution phase
I agree that this convention is quite cryptic but in theory this ini file should be modified only by developers, and I will update the documentation explaining how this coding work as soon as I am sure about those values

pattacini: Agreed! Thanks for the exhaustive answer on this 👍

dariopasquali: Hi @Nicogene and @pattacini, thank you for your work!

From a user point of view, I'm also a bit confused with this encoding:

- It has always been in **hexadecimal**. I agree that a decimal encoding is more intuitive. However, I would add a check on the percentage values to bound them between 0 and 100%. For example, if a developer, that did not read the new documentation, put a 70h (the h or p is transparent) or simply use an old  `.ini` file, this will break the eyelids moving them to position 700%

- From what you write `00p = 0% opening = closed` and `10p = 100% opening`. if this is true, `neu, sad, cun, sur` emotions are going to have **closed eyelids**, which is not good.

dariopasquali:
Hi!
I also notice that you did not change the `setRaw` method.
So, if I send `set raw S00` to the new emotionInterface with RFE board, nothing is going to happen.

I understand that updating that method keeping the retro compatibility could be a bit cryptic.
In particular, regarding the different encoding of the control messages.

However, it would be really helpful to have an interface that works both with old and new robots with the old hexadecimal percentage of opening.
You could just keep the current `writePort(cmd.c_str());` that works with old robots
And add the eyelids control for the new RFE boards as you did in the `setEyelids` method, including the conversion from hex to dec.

Nicogene:
Hi @dariopasquali ,

A check on percentage bound has been set, if greater than 100% it is set to 100%.
Moreover the inversion of the behaviour in the emotions for the closing/opening of the eyelids has been fixed, thanks.
I checked and the emotions are now the same of old robots like `iCubGenova01`.

Finally this https://github.com/robotology/icub-main/pull/668/commits/024d99706a2c41c412b074adf27d0d72a843b9f2 added the support of the new eyelids also for the `setRaw` command.

This command was intended to write directly the array of chars to the serial, for the new eyelids is no longer possible, but with a I managed to mimic this behaviour

Nicogene:
It has been tested the sequence `neu -> hap -> neu -> evi`

|  `iCubGenova01` ⚫  |  `iCubGenova02` 🟣 |
|:-:|:-:|
| ezgif com-video-to-gif(2) | ezgif com-video-to-gif(3) |

PR #673, opened 2020-07-21
SW modifications for the ukit of the upperarm cover

Nicogene:
Here is the ukit:
https://icub-tech-iit.github.io/documentation/upgrade_kits/upperarm_cover_icub2_5/support/

It is wip, I have not clear how many things will be includede in this PR

For now I just add skinGui left/right_arm2_7 ini and generators.

Following @maggia80 suggestion I edited the `lower` part of the generators
From 
immagine

to

immagine

Probably we have to check with the skin gui a possible simmetry issue


Nicogene:
The `iCubSkinGui` now appears like this

immagine


Probably one of the two patches has to be rotated

PR #679, opened 2020-09-08
FirmwareUpdater - configurable IPs

davidetome:
This PR allows configuring the IP's of the EMS boards through the `FirmwareUpdater` as per the request in https://github.com/robotology/QA/issues/423.

### DIFF
--- a/src/tools/FirmwareUpdater/firmwareupdatercore.cpp
+++ b/src/tools/FirmwareUpdater/firmwareupdatercore.cpp
@@ -10,6 +10,42 @@ static void updateProgressCallback(float fraction)
     self->updateProgress(fraction);
 }
 
+// 09/2020 davide.tome@iit.it - address and port configurable in firmwareupdater.ini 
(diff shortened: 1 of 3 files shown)


pattacini (approved):
Fine with me as we've tested it.
Probably, it's also worth awaiting @marcoaccame 's review.

Nicogene (approved):
Same from my side, maybe the debug messages can be removed

davidetome:
> Same from my side, maybe the debug messages can be removed

I just added the message if the default ip is overwritten by the one in the config file; I think it is usefull...

pattacini:
Hey @davidetome 

Next time it's better off devoting specific PR to specific developments. In this case, the changes implementing the CLI option would have best fitted a second PR.

Anyway, let's proceed this way. I'll only update the title of this PR to reflect its content.

marcoaccame (approved):
looks good to me.

PR #690, opened 2020-11-11
Fix on icubSIM joint limits

randaz81:
icubSimulationControl: fixed check on joint limits when a position command is received.


pattacini (approved):
Thanks!

PR #703, opened 2020-12-15
embobjMotionControl: fix windows execution

Nicogene:
Under Windows using `reserve` without any `resize` throws a "out of range" exception. 
This is due to the fact that after that `reserve` the `capacity` is `nj`, the size is `0`.
I am surprised that this code is allowed under linux, but maybe I am missing something.

BTW this code is tested on windows and it works


pattacini (approved):
Thanks @Nicogene 👍🏻

PR #705, opened 2020-12-16
Added command GOTO <pose> in Actions Rendering Engine

Iaxama:
With this PR the ARE accepts a new command which allows to move the robot to a specific pose. The possible poses can be added in the corresponding ini file and referenced using a tag.


app/yarpmanager/conf/cluster-config.xml
```diff
@@ -1,11 +1,17 @@
...
 this is usually the value of the DISPLAY variable when logged on -->
-<node display=":0.0">icubsrv</node>
-<node>pc104</node>
+<node>icub-head</node>
+<node>icub-zynq</node>
+<node display=":0.0">icub23</node>
+<node display=":0.0">icub24</node>
+<node>icub-spinnlink</node>
+<node>icub-blade</node>
+<node>icub-cuda</node>
```
Nicogene, line 14: I suppose that this file has been added by mistake, the `cluster-config.xml` should be a template to configure for each cluster

Nicogene (changes requested):

pattacini:
Hi @Iaxama 

Although I can clearly see the effort you put into this work, I feel this PR proposes changes that are rather off-topic in the ARE context.

ARE was born to provide a sort of high-level wrapper capable of executing **coordinated actions in the Cartesian space**, meaning actions that involve different parts (arms, head, fingers) in the operational domain in a given sequential way.

It seems to me that the control of parts in the joint space can be achieved using other standard tools we usually employ as, for example, the combination of [**`ctpService`**][1] and **bash scripts** (take a look at the bunch of scripts available in [`robotology/funny-things/app/scripts/shells`][2]). This combination works great and lies at the heart of the infrastructure we use to run demos for experiments as well as TV shows.

I've tried above to detail out the main reason why I believe this PR doesn't fit with the current architecture, but please consider also the points below:
- ARE has been growing like a monolithic beast and the central idea is to freeze it and start off following a very different development roadmap, more modular, more effective.
- Managing yet another database of actions within ARE turns to be quite overlapping and cumbersome (yep, there exists already one).
- Better off resorting to ARE only for complex sequential actions and use other tools/modules for simpler movements. Here's a list of components belonging to the `icub-main` bundle:
  - `ctpService`
  - `cartesianInterfaceControl`
  - `wholeBodyPlayer` 

[1]: https://robotology.github.io/robotology-documentation/doc/html/group__ctpService.html
[2]: https://github.com/robotology/funny-things/tree/master/app/scripts/shells

pattacini:
A few other minor points @Iaxama:
- you may consider signing up the [CLA][1] to be ready for further contributions.
- we tend to use the term `pose` for 6D goals of the end-effector, whereas we tend to use the `term` configuration to identify the joints' values.

[1]: https://github.com/robotology/icub-main/pull/705#issuecomment-746006070

pattacini:
Somehow related to https://github.com/robotology/QA/discussions/454.

PR #711, opened 2021-01-07
[SWIG] Add support for std::deque of yarp::dev::IControlLimits*

### LINKED ISSUE #710: Use alignJointsBounds from SWIG
Classes in namespace `iCub::iKin` implement a method

```
alignJointsBounds(const std::deque< yarp::dev::IControlLimits*>& lim)	
```

that alignes the limb joints bounds with the current values set aboard the iCub.

At the moment, it seems not possible to call this methods from SWIG target languages, e.g. Python, because there is no clear way - at least I cannot find any in the web - to construct something similar to a `std::deque< yarp::dev::IControlLimits*>` using the facilities provided by the target language.

E.g. in python I tried with 

```
limb.alignJointsBounds([limits])
```

where `limb` is a an instance of a class derived from ` iCub::iKin::iKinLimb`, in this case `iCubFinger`, and `[limits]` is a list containing an instance of `yarp::dev::IControlLimits` but the execution hangs with the following error

```
TypeError: in method 'iCubFinger_alignJointsBounds', 
argument 2 of type 'std::deque< yarp::dev::IControlLimits *,
std::allocator< yarp::dev::IControlLimits * > > const &'
```

After reading the SWIG [documentation](http://www.swig.org/Doc1.3/Library.html#Library_nn15) and the `icub-main` SWIG bindings file [icub.i](https://github.com/robotology/icub-main/blob/177938fbd80d133ab4157ae864c7785149105e04/bindings/icub.i) it seems that it is necessary to add explicit definition of templates that the final user needs

E.g.

https://github.com/robotology/icub-main/blob/177938fbd80d133ab4157ae864c7785149105e04/bindings/icub.i#L159-L165

I was able to use `alignJointBounds` after adding the following in `icub.i`

```
%template(dequeIControlLimitsPtr) std::deque<yarp::dev::IControlLimits*>;
```

with the following code (at least in Python)

```
from icub import dequeIControlLimitsPtr
limits_deque = dequeIControlLimitsPtr()
limits_deque.append(limits)
limb.alignJointsBounds(limits_deque)
```

If this approach sounds reasonable, I can open a PR that adds the missing declaration in `icub.i`. If the name `dequeIControlLimitsPtr` does not seem right, we can discuss a different naming convention.

xEnVrE:
As per the title, this PR add SWIG support for `std::deque<yarp::dev::IControlLimits*>` that is used, among others, in `iCub::iKin` derived classes implementing the method `alignJointsBounds`.

Fixes #710

### DIFF
--- a/bindings/icub.i
+++ b/bindings/icub.i
@@ -164,6 +164,7 @@ typedef yarp::os::BufferedPort<iCub::skinDynLib::dynContactList> BufferedPortDyn
 %template(TypedReaderCallbackDynContactList) yarp::os::TypedReaderCallback<iCub::skinDynLib::dynContactList>;
 %template(BufferedPortDynContactList) yarp::os::BufferedPort<iCub::skinDynLib::dynContactList>;
 
+%template(dequeIControlLimitsPtr) std::deque<yarp::dev::IControlLimits*>;
 
 bool init();
 


pattacini (approved):

xEnVrE:
Thank you @pattacini

Issue #712, opened 2021-01-07
embObjMotionControl: issue in CONTROL configuration parsing


lrapetti:
While changing the parameters in the `CONTROL` group configuration I found the following issue which might lead to problem not easy to spot (as it was in my experience).

The `CONTROL` group usually is something like:
```xml
<group name="CONTROLS">
        <param name="positionControl">          POS_PID_DEFAULT_1     POS_PID_DEFAULT_2 </param>
        <param name="velocityControl">          POS_PID_DEFAULT_1     POS_PID_DEFAULT_2 </param>
        <param name="mixedControl">             POS_PID_DEFAULT_1     POS_PID_DEFAULT_2 </param>
        <param name="torqueControl">            TRQ_PID_DEFAULT_1     TRQ_PID_DEFAULT_2 </param>
        <param name="currentPid">               2FOC_CUR_CONTROL_1    2FOC_CUR_CONTROL_2 </param>
        <param name="speedPid">                 2FOC_VEL_CONTROL_1    2FOC_VEL_CONTROL_2 </param>
</group>
```
this means that we are defining a set of gains for each joint (each column) as defined in the given `POS_PID_DEFAULT_{i}` group.
The `POS_PID_DEFAULT_{i}` group is something like:
```xml
group name="POS_PID_DEFAULT_{i}">
        <param name="controlLaw">           minjerk             </param>
        <param name="outputType">           pwm                 </param>
        <param name="fbkControlUnits">      metric_units        </param>
        <param name="outputControlUnits">   machine_units       </param>
        <param name="kp">                   711.11      1066.66 </param>
        <param name="kd">                   0.00        0.00 </param>
        <param name="ki">                   7111.09     10666.64 </param>
        <param name="maxOutput">            8000        8000 </param>
        <param name="maxInt">               200         200 </param>
        <param name="stictionUp">           0           0 </param>
        <param name="stictionDown">         0           0 </param>
        <param name="kff">                  0           0 </param>
</group>
```
as you can see, it contains the gains for both the two joints and depending on which column of the `CONTROLS` group this configuration is used, the gains of the corresponding column will be used.

Up to now everything is fine. However, you can see that some of the configuration parameters (e.g. `controlLaw`, `outputType`, `fbkControlUnits`) are defined only once for all the joints. What happen if the value of those parameter is different in the different columns?
[...]

PR #716, opened 2021-02-03
Warning message for position control PID when config not consistent

AlexAntn:
This pull request introduces a warning message when the configuration of position control PID is not consistent across the different joints, resulting in wrong values for the PIDs (mismatching between `machine_units` and `metric_units`).

This PR addresses [this issue](https://github.com/robotology/icub-main/issues/712).

@randaz81 : This implementation covers **only position control**, since the configuration for other types of control is slightly different. Should this be implemented for all different types of control?

### DIFF
--- a/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
@@ -2327,7 +2327,7 @@ void JointsSet::dumpdata(void)
 
 }
 
-bool Parser::checkJointTypes(PidInfo *pids, std::string pid_type)
+bool Parser::checkJointTypes(PidInfo *pids, const std::string pid_type)
 {
     //Here i would check that all joints have same type units in order to create pid_type helper with correct factor.
 
--- a/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
@@ -2362,3 +2362,4 @@ bool Parser::checkJointTypes(PidInfo *pids, std::string pid_type)
     }
     return true;
 }
+
--- a/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
@@ -242,39 +242,7 @@ bool Parser::parseSelectedCurrentPid(yarp::os::Searchable &config, bool pidisMan
         }
     }
 
-    //Here i would check that all joints have same type units in order to create current helper with correct factor.
-
-    //get first joint with enabled current
-    int firstjoint = -1;
-    for(int i=0; i<_njoints; i++)
-    {
-        if(pids[i].enabled)
-        {
-            firstjoint = i;
-            break;
-        }
-    }
-
-    if(firstjoint==-1)
-    {
-        // no joint has current enabed
-        return true;
-    }
-
-    for(int i=firstjoint+1; i<_njoints; i++)
-    {
-        if(pids[i].enabled)
-        {
-            if(pids[firstjoint].fbk_PidUnits != pids[i].fbk_PidUnits ||
-               pids[firstjoint].out_PidUnits != pids[i].out_PidUnits)
-            {
-                yError() << "embObjMC BOARD " << _boardname << "all joints with CURRENT enabled should have same controlunits type. Joint " << firstjoint << " differs from joint " << i;
-                return false;
-            }
-        }
-    }
-
-    return true;
+    return checkJointTypes(pids, "CURRENT");
 }
 
 bool Parser::parseSelectedSpeedPid(yarp::os::Searchable &config, bool pidisMandatory, PidInfo *pids) // OK
@@ -339,39 +307,7 @@ bool Parser::parseSelectedSpeedPid(yarp::os::Searchable &config, bool pidisManda
         }
     }
 
-    //Here i would check that all joints have same type units in order to create speed helper with correct factor.
-
-    //get first joint with enabled speed
-    int firstjoint = -1;
-    for(int i=0; i<_njoints; i++)
-    {
-        if(pids[i].enabled)
-        {
-            firstjoint = i;
-            break;
-        }
-    }
-
-    if(firstjoint==-1)
-    {
-        // no joint has speed enabed
-        return true;
-    }
-
-    for(int i=firstjoint+1; i<_njoints; i++)
-    {
-        if(pids[i].enabled)
-        {
-            if(pids[firstjoint].fbk_PidUnits != pids[i].fbk_PidUnits ||
-               pids[firstjoint].out_PidUnits != pids[i].out_PidUnits)
-            {
-                yError() << "embObjMC BOARD " << _boardname << "all joints with SPEED enabled should have same controlunits type. Joint " << firstjoint << " differs from joint " << i;
-                return false;
-            }
-        }
-    }
-
-    return true;
+    return checkJointTypes(pids, "SPEED");
 }
 
 bool Parser::parseSelectedPositionControl(yarp::os::Searchable &config) // OK
@@ -1197,9 +1133,6 @@ bool Parser::parsePid_torque_outVel(Bottle &b_pid, string controlLaw)
     return true;
 }
 
-
-
-
 bool Parser::getCorrectPidForEachJoint(PidInfo *ppids/*, PidInfo *vpids*/, TrqPidInfo *tpids)
 {
     Pid_Algorithm *minjerkAlgo_ptr = NULL;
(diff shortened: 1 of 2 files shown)


pattacini:
cc @lrapetti

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -1278,6 +1278,35 @@ bool Parser::getCorrectPidForEachJoint(PidInfo *ppids/*, PidInfo *vpids*/, TrqPi
...
 
+    //Here i would check that all joints have same type units in order to create position helper with correct factor.
+
+    //get first joint with enabled position
+    firstjoint = -1;
+    for(int i=0; i<_njoints; i++)
+    {
+        if(ppids[i].enabled)
+            firstjoint = i;
+    }
```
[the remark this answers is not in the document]
pattacini: We might want to fix here and there then as I can see how this logic mismatch could potentially affect the cycle soon afterward.

pattacini (commented):
A few hints.

lrapetti:
Thanks @AlexAntn for taking care.

> @randaz81 : This implementation covers **only position control**, since the configuration for other types of control is slightly different. Should this be implemented for all different types of control?

If I can give my opinion regarding this point, in general I think this needs to be implemented for all the control types. For example, [my issue](https://github.com/robotology/icub-main/issues/712) was opened after I experienced that problem in torque control mode.

pattacini:
>If I can give my opinion regarding this point, in general I think this needs to be implemented for all the control types. For example, my issue was opened after I experienced that problem in torque control mode.

But the case of torque PID should have been already covered, if I got it right (cc @AlexAntn).
Maybe because of the logic I pointed out [here][1] the warning wasn't generated.

https://github.com/robotology/icub-main/pull/716#discussion_r569778460

lrapetti:
> > If I can give my opinion regarding this point, in general I think this needs to be implemented for all the control types. For example, my issue was opened after I experienced that problem in torque control mode.
> 
> But the case of torque PID should have been already covered if I got it right (cc @AlexAntn).
> Maybe because of the logic I pointed out [here](https://github.com/robotology/icub-main/pull/716#discussion_r569778460) the warning wasn't generated.

Should it be a `Warning` or should it return an `Error`? I did't notice anything, but it could be possible I was missing the `Warning`.

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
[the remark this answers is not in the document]
AlexAntn: In fact, looking at the logic, the `firstjoint` will in fact be `lastjoint`. I will edit the code with these changes and push

AlexAntn:
> > > If I can give my opinion regarding this point, in general I think this needs to be implemented for all the control types. For example, my issue was opened after I experienced that problem in torque control mode.
> > 
> > 
> > But the case of torque PID should have been already covered if I got it right (cc @AlexAntn).
> > Maybe because of the logic I pointed out [here](https://github.com/robotology/icub-main/pull/716#discussion_r569778460) the warning wasn't generated.
> 
> Should it be a `Warning` or should it return an `Error`? I did't notice anything, but it could be possible I was missing the `Warning`.

The message is an `error` (for both torque and position control), not a `warning`, so my guess is it would be very visible if it showed up. As @pattacini mentioned, there is an error in the logic of the code (regarding the `firstjoint`), which would lead to no message being printed in any case. I will fix it and push.

pattacini:
>which would lead to no message being printed in any case

In some cases, not all, but I think fixing this will make sure to deal with the correct joint number.

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -85,7 +85,7 @@ bool Parser::parsePids(yarp::os::Searchable &config, PidInfo *ppids/*, PidInfo *
     // come specificato in _currentControlLaw
     if(!parseSelectedCurrentPid(config, lowLevPidisMandatory, cpids)) // OK
         return false;
-    // legge i pid di velocit� per ciascun motore 
+    // legge i pid di velocita per ciascun motore
```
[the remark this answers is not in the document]
AlexAntn: I "fixed" it because in my vscode it was unable to process the accent on the letter

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -242,6 +242,38 @@ bool Parser::parseSelectedCurrentPid(yarp::os::Searchable &config, bool pidisMan
...
+        {
+            if(pids[firstjoint].fbk_PidUnits != pids[i].fbk_PidUnits ||
+               pids[firstjoint].out_PidUnits != pids[i].out_PidUnits)
+            {
+                yError() << "embObjMC BOARD " << _boardname << "all joints with CURRENT enabled should have same controlunits type. Joint " << firstjoint << " differs from joint " << i;
+                return false;
+            }
+        }
+    }
+
```
AlexAntn, line 276: Here I introduced an error message for current control type

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -307,6 +339,38 @@ bool Parser::parseSelectedSpeedPid(yarp::os::Searchable &config, bool pidisManda
...
+        {
+            if(pids[firstjoint].fbk_PidUnits != pids[i].fbk_PidUnits ||
+               pids[firstjoint].out_PidUnits != pids[i].out_PidUnits)
+            {
+                yError() << "embObjMC BOARD " << _boardname << "all joints with SPEED enabled should have same controlunits type. Joint " << firstjoint << " differs from joint " << i;
+                return false;
+            }
+        }
+    }
+
```
AlexAntn, line 373: here I introduced a similar message for speed control type

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -1246,8 +1310,6 @@ bool Parser::getCorrectPidForEachJoint(PidInfo *ppids/*, PidInfo *vpids*/, TrqPi
         //eomc_ctrl_out_type_vel = 2,
         //eomc_ctrl_out_type_cur = 3
 
-    return true;
-
```
AlexAntn, line 1250: We noticed there was this `return` before any of the error messages were printed, was this intentional or a small bug, @randaz81 ?

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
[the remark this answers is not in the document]
pattacini: You may consider translating the whole comment into English.

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -1256,7 +1318,10 @@ bool Parser::getCorrectPidForEachJoint(PidInfo *ppids/*, PidInfo *vpids*/, TrqPi
     for(int i=0; i<_njoints; i++)
     {
         if(tpids[i].enabled)
+        {
             firstjoint = i;
+            break;
+        }
```
AlexAntn, line 1324: Introduced a break on the `firstjoint` loop, so now it correctly gets the first joint (instead of the last enabled one)

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -1272,7 +1337,39 @@ bool Parser::getCorrectPidForEachJoint(PidInfo *ppids/*, PidInfo *vpids*/, TrqPi
...
+    }
+
+    for(int i=firstjoint+1; i<_njoints; i++)
+    {
+        if(ppids[i].enabled)
+        {
+            if(ppids[firstjoint].fbk_PidUnits != ppids[i].fbk_PidUnits ||
+               ppids[firstjoint].out_PidUnits != ppids[i].out_PidUnits)
+            {
+                yError() << "embObjMC BOARD " << _boardname << "all joints with POSITION enabled should have same controlunits type. Joint " << firstjoint << " differs from joint " << i;
```
AlexAntn, line 1372: finally, added the error message for position control (not just torque). Also, made the origin of the messages more evident by putting in caps (e.g.: `POSITION`) the type of control that is reporting the error

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
pattacini: Because of this, there was no hope to see error messages popping out.

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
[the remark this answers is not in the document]
AlexAntn: maybe it is better to leave that for another PR to not mix up with a lot of changes - then I can translate the Italian comments I think

pattacini:
@randaz81 @lrapetti the PR is somewhat ready for your reviews.

pattacini (approved):

randaz81:
Nice fix!
Just an additional question: at first glance it seems to me that the code of that performs the check is identical in all the three different cases (only the output string changes?). If this is the case, maybe a more compact solution to avoid code duplication (if in the future additional control blocks are added) could be to have a call to a common method that takes in input the pid and its name (string) and returns true/false to the upper level? Just an idea..

pattacini:
I think it's a valuable feedback @randaz81 👍🏻

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -2385,4 +2482,3 @@ void JointsSet::dumpdata(void)
     cout << " param1="<< cfg.constraints.param1 << " param2=" << cfg.constraints.param2 << endl;
 
 }
-
```
lrapetti, line 2388: Nitpick: usually is good practice to leave an empty line at the end of the file

lrapetti (approved):

AlexAntn:
As @randaz81 suggested, I put the print into a function that is now called (instead of having the same code spread all over).

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -370,6 +370,7 @@ class Parser
     bool getCorrectPidForEachJoint(PidInfo *ppids/*, PidInfo *vpids*/, TrqPidInfo *tpids);
     bool parsePidUnitsType(yarp::os::Bottle &bPid, yarp::dev::PidFeedbackUnitsEnum  &fbk_pidunits, yarp::dev::PidOutputUnitsEnum& out_pidunits);
 
+    bool checkJointTypes(PidInfo *pids, std::string pid_type);
```
pattacini, line 373: This ought to be `const std::string& pid_type`.

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -2386,3 +2327,39 @@ void JointsSet::dumpdata(void)
 
 }
 
+bool Parser::checkJointTypes(PidInfo *pids, std::string pid_type)
```
pattacini, line 2330: This ought to be `const std::string& pid_type`.

pattacini (changes requested):
Minor comments

src/libraries/icubmod/embObjMotionControl/eomcParser.h
AlexAntn: done!

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
AlexAntn: done!

src/libraries/icubmod/embObjMotionControl/eomcParser.h
pattacini: It's important the `&` as it needs to be a reference.

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
pattacini: Same comment here.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
AlexAntn: oops, fixed! (this time for real)

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
AlexAntn: fixed!

pattacini (approved):

pattacini:
Thanks @AlexAntn 👍🏻 
I'll merge the PR once checks are passed.

lrapetti:
Thanks all!

Issue #718, opened 2021-02-23
embobjMotionControl fails to open after adding the consistency check of torque PIDs


Nicogene:
This issue is a regression of #716.

Here is my xml file

```xml
  
    <group name="CONTROLS">
        <param name="positionControl">          POS_PID_DEFAULT     POS_PID_DEFAULT     POS_PID_DEFAULT     POS_PID_DEFAULT     </param>
        <param name="velocityControl">          POS_PID_DEFAULT     POS_PID_DEFAULT     POS_PID_DEFAULT     POS_PID_DEFAULT     </param>
        <param name="mixedControl">             POS_PID_DEFAULT     POS_PID_DEFAULT     POS_PID_DEFAULT     POS_PID_DEFAULT     </param>
        <param name="torqueControl">            TRQ_PID_DEFAULT     TRQ_PID_DEFAULT     TRQ_PID_DEFAULT     TRQ_PID_DEFAULT     </param>
        <param name="currentPid">               2FOC_CUR_CONTROL    2FOC_CUR_CONTROL    2FOC_CUR_CONTROL    2FOC_CUR_CONTROL    </param>
        <param name="speedPid">                 2FOC_VEL_CONTROL    2FOC_VEL_CONTROL    2FOC_VEL_CONTROL    2FOC_VEL_CONTROL    </param>
    </group>



    <!-- other default PIDs: begin -->  
    
    <group name="TRQ_PID_DEFAULT">
        <param name="controlLaw">           torque              </param>
        <param name="outputType">           pwm                 </param>
        <param name="fbkControlUnits">      metric_units        </param>
        <param name="outputControlUnits">   machine_units       </param>
        <param name="kp">                   200         200         250         300     </param>
        <param name="kd">                   0           0           0           0       </param>
        <param name="ki">                   0           0           0           0       </param>
        <param name="maxOutput">            8000        8000        8000        8000    </param>
        <param name="maxInt">               500         500         500         500     </param>
        <param name="ko">                   0           0           0           0       </param>
        <param name="stictionUp">           0           0           0           0       </param>
        <param name="stictionDown">         0           0           0           0       </param>
        <param name="kff">                  1           1           1           1       </param>
        <param name="kbemf">                0           0           0           0       </param>
        <param name="filterType">           0           0           0           0       </param>
        <param name="ktau">                 0           0           0           0       </param>
    </group>
```

Here is the error I get
```
[...]

PR #719, opened 2021-02-24
fixed the checkJointTypes function for the case of TORQUE control

AlexAntn:
This PR fixes issue #718 .

Since we have an array of PidInfo / TrqPidInfo, we can't rely on class inheritance to query the parameters of each PID ([an array of Derived is not a kind-of array of Base][1]). As such, we need to check if the PID type is torque and handle it differently.

[1]: https://www.cs.technion.ac.il/users/yechiel/c++-faq/array-derived-vs-base.html

### DIFF
--- a/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
@@ -2364,3 +2364,4 @@ bool Parser::checkJointTypes(PidInfo *pids, const std::string &pid_type)
     return true;
 }
 
+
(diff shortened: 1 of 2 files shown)


pattacini (changes requested):
I can see a verbatim copy'n'paste of lots of code lines we could thus refactor and made much more compact.

pattacini:
@AlexAntn we can get away w/ less by refactoring the main routine and make it call 2 sub-routines that implement each the body of the 2 for loops.

AlexAntn:
Ìn the last commit to this issue I have refactored the loops in order to reduce duplicated code. I have also changed how we obtain the `firstjoint` in order to have a single loop.

These changes have been tested directly on the robot and are working (`yarprobotinterface` started with no issues)

pattacini (approved):

pattacini:
LGTM, thanks a lot @AlexAntn 🚀

pattacini:
I'll wait for the CI to complete and the squash and merge the PR.

@Nicogene in the meantime, you may be willing to review it as well.

Nicogene (approved):
LGTM

Nicogene:
I think we can squash all the commit in just one

pattacini:
>I think we can squash all the commit in just one

Don't do that please, otherwise CI will start over from scratch!

I can simply use the corresponding button from the GH GUI instead.

PR #726, opened 2021-03-08
wholeBodyPlayer: fix limits/security checks

Nicogene:
The latest joint of the wrist is 6 not 4, and the check of the joint limits has to be skipped for the hands since the encoders are not reliable

TO BE TESTED


pattacini:
>check of the joint limits has to be skipped for the hands since the encoders are not reliable

what do you mean?

Nicogene:
Replaying a trajectory I recorded on the same robot I get tons of errors on hands joints:

```
```

pattacini:
>Maybe the joints limit for the hand are not configured correctly in the XML files

Exactly, that might be the point.

pattacini:
Any update on this @Nicogene ?

Nicogene:
We discovered that there are other synchronization issues, but we will address them the next PI, closing for now

PR #741, opened 2021-04-08
Fix `ARE` check on action type

### LINKED ISSUE #740: actionsRenderingEngine: calib table stopped working
Today I tested the `calib table` command with `actionRenderingEngine` of the latest tag (`v1.19.2`) and it never detect a contact, I put some prints and I found out that `contactDetectionOn` was always false and it is used while checking the table height:
https://github.com/robotology/icub-main/blob/89971ccd0eda518f2b3c972d1b805c2f44c411fc/src/libraries/actionPrimitives/src/actionPrimitives.cpp#L2212-L2232

Then I searched who should set that flag to true and I found out that this check:
https://github.com/robotology/icub-main/blob/89971ccd0eda518f2b3c972d1b805c2f44c411fc/src/modules/actionsRenderingEngine/src/MotorThread.cpp#L2538-L2539

is always false because:

```
action: A2_PN4iCub6action16ActionPrimitivesE
Primitive: N4iCub6action22ActionPrimitivesLayer2E

```

pattacini:
This PR fixes #740.

I took the chance to analyze the problem by using the following snippet:
```c++
#include <cstdlib>
#include <typeinfo>
#include <string>
#include <iostream>

#include <iCub/action/actionPrimitives.h>

void test(const int i, const bool cond) {
    std::cout << "test " << i << ": "
              << (cond ? "passed ✅" : "failed ❌")
              << std::endl;
}

int main() {
    iCub::action::ActionPrimitives* action = new iCub::action::ActionPrimitivesLayer2();

    test(1, typeid(action).name() == typeid(iCub::action::ActionPrimitivesLayer2).name());
    test(2, typeid(*action).name() == typeid(iCub::action::ActionPrimitivesLayer2).name());
    test(3, dynamic_cast<iCub::action::ActionPrimitivesLayer2*>(action));
    test(4, typeid(action) == typeid(iCub::action::ActionPrimitivesLayer2));
    test(5, typeid(*action) == typeid(iCub::action::ActionPrimitivesLayer2));

    delete action;
    return EXIT_SUCCESS;
}
```

whose results are:
```console
test 1: failed ❌
test 2: passed ✅
test 3: passed ✅
test 4: failed ❌
test 5: passed ✅
```

That said, I went for the patch used in test no. 3 as it eases the `ARE` code as well.

@Nicogene I found the use of `std::is_same<,>` along with `decltype()` a bit overkill plus I didn't manage to get it work at once.

### DIFF
--- a/src/modules/actionsRenderingEngine/src/MotorThread.cpp
+++ b/src/modules/actionsRenderingEngine/src/MotorThread.cpp
@@ -17,7 +17,6 @@
 */
 
 
-#include <typeinfo>
 #include <fstream>
 #include <sstream>
 #include <cmath>
@@ -1260,8 +1259,8 @@ bool MotorThread::threadInit()
             action[arm]->getGraspModel(graspModel[arm]);
             action[arm]->enableReachingTimeout(reachingTimeout);
 
-            if (typeid(action).name()==typeid(ActionPrimitivesLayer2).name())
-                dynamic_cast<ActionPrimitivesLayer2*>(action[arm])->setExtForceThres(extForceThresh[arm]);
+            if (auto* p = dynamic_cast<ActionPrimitivesLayer2*>(action[arm]))
+                p->setExtForceThres(extForceThresh[arm]);
 
             deque<string> q=action[arm]->getHandSeqList();
             yInfo("***** List of available %s hand sequence keys:",arm_name[arm].c_str());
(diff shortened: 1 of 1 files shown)


Nicogene (approved):
Super thanks!

pattacini:
@Nicogene the final test would be to make it run on the real robot, also because it remains unclear how `ARE` was able to be working as expected in the past 🤔

Issue #768, opened 2021-10-18
When a joint goes in HF, errors like `positionMoveRaw: skipping command` flood the logger


S-Dafarra:
### Bug description

When writing simple applications to control the robot, we are used to sending commands to joints without checking the current control mode. See for example https://github.com/robotology/icub-tutorials/blob/23ac25487d5a872030b82b38a28eb44b2cb3bda3/src/motorControlBasic/tutorial_arm.cpp#L129

This is because we usually set the desired control mode only once, during the initialization phase, and during the control loop we simply get some measurement, and we set the references to the joints.

It may happen though, that some joint goes in HF after the startup. This happens quite easily with the hands. After that, the logger gets quickly filled up with messages like:
```
<ERROR> velocityMoveRaw: skipping command because  BOARD left_arm-eb26-j12_15 (IP 10.0.1.26)   joint  2  is not in VOCAB_CM_VELOCITY mode
```
Since the top level application usually does not check for the control mode at each control loop, it keeps sending references to the joint, producing errors like the above. Since the control loop is usually at 100Hz, this will saturate the logger quite easily, losing the possibility to check other errors.  

cc @pattacini @marcoaccame 

### Steps to reproduce

It should be enough to run the tutorial https://github.com/robotology/icub-tutorials/tree/23ac25487d5a872030b82b38a28eb44b2cb3bda3/src/motorControlBasic on the robot and change the control mode of the controlled joints to something different from ``Position`` and ``Idle``.

### Expected behavior

The errors of https://github.com/robotology/icub-main/blob/601af527248f16c73e3d353bf9bdac0fa47b3d9c/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp#L2224, https://github.com/robotology/icub-main/blob/601af527248f16c73e3d353bf9bdac0fa47b3d9c/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp#L1897, and https://github.com/robotology/icub-main/blob/601af527248f16c73e3d353bf9bdac0fa47b3d9c/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp#L4106

should be limited in time, eventually sending the error only once, and sending it again when the control mode changes.

pattacini:
This is not a bug, but rather more a feature request, I'd say.

In my opinion, the high-level SW should be responsible somehow for checking the status of the boards. We do this for example in the Cartesian and Gaze controller.

[...]

PR #770, opened 2021-11-16
Add message downsampler to avoid logger flooding when changing control mode

mfussi66:
This PR aims to fix the issue: https://github.com/robotology/icub-main/issues/768
To downsample the error messages rate, the logic of a simple Finite State Machine is implemented inside a `yarp::os::Timer`. The timer runs at 100Hz. If they exceed a threshold of 5 events, the FSM prints the number of error events occurred each second. 
The proposed solution was tested on a desk setup with one joint, addressing the three  types of error events mentioned in the issue.

The videos below show the results on the setup.

On the top left there's the `Yarpmotorgui`, where I change the control mode. On the top right there's the program `tutorial_arm` that sends the requests, and triggers the errors in the `yarprobotinterface` if the control mode is not the expected one. On the bottom there is the `yarprobotinterface`.
 - From Position to Velocity

https://user-images.githubusercontent.com/38140169/141956329-92ffd840-57a6-4640-8e18-a4eefabee91f.mp4

- From Velocity to Position

https://user-images.githubusercontent.com/38140169/141957102-11277cc6-3883-4173-be1c-03c068cf45db.mp4

- From Position Direct to Velocity

https://user-images.githubusercontent.com/38140169/141957131-3839e8ea-926c-4463-b2c2-63228d5edaa4.mp4

### DIFF
--- a/src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
+++ b/src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
@@ -6,11 +6,9 @@
 
 namespace mced {
     
-    
     mcEventDownsampler::mcEventDownsampler() 
     {
         mutex = new std::mutex();
-        // mutex ctor   
     };
     
     mcEventDownsampler::~mcEventDownsampler()
@@ -23,36 +21,40 @@ namespace mced {
             delete timer;
             timer = nullptr;
         }
-        // idem per le altre cose    
-        
-        // mutex dtor       
+
+        delete mutex;
     }
     
     bool mcEventDownsampler::canprint()
     {
-        bool cp = true;
         mutex->lock();
-        // activate something.
+        
         counter++;
-
-        if(counter - latch_1 > 5)
-        {
-            cp = false;
-        } else {
-            cp = !isdownsampling;
-        }
+        size_t diff = counter - latch_1;
+        bool cp = !isdownsampling;
 
         mutex->unlock();
-        
-        return cp;
+
+        if(diff > 5)
+        {
+            return false;
+        } 
+        else
+        {
+            return cp;
+        }
     }
     
     bool mcEventDownsampler::stop()
     {
-        // controlli su valore di timer non null, se e'attivo etc....
-        
-        timer->stop();
-        return true;
+
+        if(nullptr != timer)
+        {
+            timer->stop();
+            return true;
+        }
+
+        return false;
     }
     
     bool mcEventDownsampler::start(const Config &config)
@@ -60,33 +62,29 @@ namespace mced {
         if(nullptr != timer)
         {
             stop(); 
-            // delete timer
         }
+
         expire_time = yarp::os::Time::now();        
-        yarp::os::TimerSettings ts(0.01);
-        timer = new yarp::os::Timer(ts, &mcEventDownsampler::step, this, true);
+        yarp::os::TimerSettings ts(config.period);
+        timer = new yarp::os::Timer(ts, &mcEventDownsampler::step, this, true, mutex);
 
-
-        timer->start();
-        // decidere dove costruire mutex. di sicuro deve essere costruito prima di timer
-        // meglio nel ctor
-        // construct timer etc
-        
-        // assegnare step() come callback al timer ....
-        
-        return true;
+        if (timer != nullptr) 
+        {
+            timer->start();
+            return true;
+        }
+        else
+        {
+            return false;
+        }
     }
     
     bool mcEventDownsampler::step(const yarp::os::YarpTimerEvent &event)
     {
-        // qui dentro, deve esserci la logica della fsm e ci deve essere accesso ai dati privati della class Ticker
-        mutex->lock();        
-        //fsmstep(fsm);
-
         if (yarp::os::Time::now() - expire_time >= 1)
         {
             expire_time = yarp::os::Time::now();
-
+            
             isdownsampling = (counter - latch_1 > 5);
 
             if (isdownsampling)
@@ -94,12 +92,8 @@ namespace mced {
                 printreport();
                 latch_2 = counter;
             }
-            latch_1 = counter;
+            latch_1 = counter;            
         }
-
-        // genera ....
-        
-        mutex->unlock();
         return true;
     }
 
@@ -107,10 +101,5 @@ namespace mced {
     {
         yError() <<  "Detected " << (counter - latch_2) << " events on aggregate since the last message";
     }
-    
-    void mcEventDownsampler::fsmstep(void *fsm) {}
-        
-} // xxx
-
-
-// - end-of-file (leave a blank line after)----------------------------------------------------------------------------
+            
+} // mced
--- a/src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
+++ b/src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
@@ -9,6 +9,7 @@ namespace mced {
     
     mcEventDownsampler::mcEventDownsampler() 
     {
+        mutex = new std::mutex();
         // mutex ctor   
     };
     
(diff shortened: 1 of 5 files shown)


pattacini:
cc @S-Dafarra

src/libraries/icubmod/embObjMotionControl/embObjMotionControl.h
```diff
@@ -256,7 +261,7 @@ class yarp::dev::embObjMotionControl:   public DeviceDriver,
     double  *_ref_accs;         // for velocity control, in position min jerk eq is used.
     double  *_encodersStamp;                    /** keep information about acquisition time for encoders read */
     bool  *checking_motiondone;                 /* flag telling if I'm already waiting for motion done */
-    #define MAX_POSITION_MOVE_INTERVAL 0.080
+    #define MAX_POSITION_MOVE_INTERVAL 0.0001
```
pattacini, line 264: What is this for?

mfussi66: I changed it to avoid printing the warning message:
"Performance warning: You are using positionMove commands at high rate (<"<<MAX_POSITION_MOVE_INTERVAL*1000.0 <<" ms)"

I will revert it to its original value.

src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
```diff
@@ -0,0 +1,105 @@
...
+                latch_2 = counter;
+            }
+            latch_1 = counter;            
+        }
+        return true;
+    }
+
+    void mcEventDownsampler::printreport()
+    {
+        yError() <<  "Detected " << (counter - latch_2) << " events on aggregate since the last message";
```
pattacini, line 102: I would grab here the chance to make the printout more meaningful by registering the downsampler as a new logging component.

See https://github.com/robotology/yarp/pull/2259 for an example.

pattacini (changes requested):
Hi @mfussi66 

Nicely done 👍🏻 

I've only a minor comment and a request for improving the printouts.

src/libraries/icubmod/embObjMotionControl/embObjMotionControl.h
pattacini: Yes thanks, revert it.

S-Dafarra:
Thanks a lot! Just as a curiosity, I have seen in the video above, you only change the control mode of a single joint. What happens if you do the same with multiple joints?

pattacini:
> Thanks a lot! Just as a curiosity, I have seen in the video above, you only change the control mode of a single joint. What happens if you do the same with multiple joints?

Good question!

In that file, there are only 3 "skipping"-like error messages and we address all of them, which are joint dependent.

@marcoaccame, how does it work with multi-joint commands? Are they unpacked and dealt with at the single joint level? If so we're ok, otherwise we're required to catch where else the error messages are yielded and manage them accordingly.

Perhaps, @mfussi66, you could quickly test the multi-joint approach as well.

mfussi66:
> Perhaps, @mfussi66, you could quickly test the multi-joint approach as well (most likely straight on the robot).

Yep, I'll check the availability of the robot and book it asap. I assume the code could be tested also on iCub2? In case iCub3 is not available.

pattacini:
> > Perhaps, @mfussi66, you could quickly test the multi-joint approach as well (most likely straight on the robot).
> 
> Yep, I'll check the availability of the robot and book it asap. I assume the code could be tested also on iCub2? In case iCub3 is not available.

Yes, but wait for @marcoaccame's input first.

S-Dafarra:
> @marcoaccame, how does it work with multi-joint commands? Are they unpacked and dealt with at the single joint level? If so we're ok, otherwise we're required to catch where else the error messages are yielded and manage them accordingly.

By experience, I think this is the case

src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
mfussi66: Added.

pattacini: Awesome!

marcoaccame:
> @marcoaccame, how does it work with multi-joint commands? Are they unpacked and dealt with at the single joint level? If so we're ok, otherwise we're required to catch where else the error messages are yielded and manage them accordingly.

So far, every call to  `velocityMoveRaw()`, `positionMoveRaw()` and `setPositionRaw()` for any joint contributes to reach the threshold. 

So, if the threshold of events is 5 and the device is for instance `left_arm-eb24-j4_7` which manages 4 joints, then one velocity move of the entire part increments 4 times and the next velocity mode exceeds 5 and triggers the threshold.

If desired, it is possible (and easy) to modify the implementation so that there is one counter per each joint.

pattacini:
Thanks @marcoaccame 👍🏻 

>So, if the threshold of events is 5 and the device is for instance left_arm-eb24-j4_7 which manages 4 joints, then one velocity move of the entire part increments 4 times and the next velocity mode exceeds 5 and triggers the threshold.

That's how it is supposed to work, so far so good 👍🏻 

>If desired, it is possible (and easy) to modify the implementation so that there is one counter per each joint.

We can certainly make things more general and configurable but I would like to go on in this state to reach a first MVP.

Therefore, I'm waiting for the CI to complete and then merge the PR.

pattacini (approved):

src/libraries/icubmod/embObjLib/mcEventDownsampler.h
```diff
@@ -0,0 +1,110 @@
...
+    {
+    public:
+
+        /**
+         * @brief Contains the configurable parameters of the event downsampler
+         * @var Config::period the period of the timer
+         */
+        struct Config
+        {
+            double period {0};
```
pattacini, line 27:
```suggestion
            double period {0.};
```

src/libraries/icubmod/embObjLib/mcEventDownsampler.h
```diff
@@ -0,0 +1,110 @@
...
+            Config() = default;
+            constexpr Config(double c) : period(c) {} 
+            bool isvalid() const { return 0 != period; }
```
pattacini, line 30:
```suggestion
            bool isvalid() const { return 0. != period; }
```

pattacini (commented):
Minor refactoring.

src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
mfussi66: Here is a screenshot showing the updated logger

image

S-Dafarra:
I would just suggest a more descriptive message. I am afraid that ``Detected 1024 events on aggregate since the last message`` might be a little cryptic if you miss the first message about the wrong control mode.

pattacini:
> I would just suggest a more descriptive message. I am afraid that `Detected 1024 events on aggregate since the last message` might be a little cryptic if you miss the first message about the wrong control mode.

We have just added the new logging component and we may rely on that.
See https://github.com/robotology/icub-main/pull/770#discussion_r750258041.

Perhaps, @mfussi66, you may call the component something like `mc.msgdownsampler.modeskipped` instead of `mc.downsampler`.

src/libraries/icubmod/embObjLib/mcEventDownsampler.cpp
```diff
@@ -0,0 +1,109 @@
+// --------------------------------------------------------------------------------------------------------------------
+// - public interface
+// --------------------------------------------------------------------------------------------------------------------
+
+#include "mcEventDownsampler.h"
+
+namespace mced {
+    
+
+    YARP_LOG_COMPONENT(MC_EVENT_DOWNSAMPLER, "mc.downsampler")
```
pattacini, line 10:
```suggestion
    YARP_LOG_COMPONENT(MC_EVENT_DOWNSAMPLER, "mc.msgdownsampler.skipped-cmd-wrong-mode")
```

cc @mfussi66

PR #772, opened 2021-11-17
Configuration of `threshold` and `subcomponent` of the message downsampler

pattacini:
This PR follows up on #771 and allows for configuring the `threshold` and the `subcomponent` name.

The latter might be required as the downsampler is in the library and could be utilized in the future also for different goals other than the MC.


mfussi66:
Looks good to me!

mfussi66 (approved):

pattacini:
Waiting for CI as I didn't build the project locally 😄

pattacini:
cc @S-Dafarra

PR #778, opened 2021-12-03
Added support for amcbldc and pmc boards in FirmwareUpdater

marcoaccame:
This PR allows the program `FirmwareUpdater` to recognize the boards `pmc` and `amcbldc` for its operativity.

For this purpose, we need to define some enum values in `icub-firmware-shared` (see the relevant [PR](https://github.com/robotology/icub-firmware-shared/pull/51)) and to avoid compilation failure we also need to update the required version of icub-firmware-shared to be 1.22.1.

A small but important change was to increase the timeout with which the `canloader` library waits for a reply from the `amcbldc` and `pmc` after it has sent them a `BOOTLOADER:board` message. The reason is that these two new boards require longer time to perform their FLASH erase.

For further details see also the following joint PR in [icub-firmware](https://github.com/robotology/icub-firmware/pull/222), where we also document about the performed tests.

      
**NOTE**: The CI will fail because we need first to merge the PR on [icub-firmware-shared](https://github.com/robotology/icub-firmware-shared/pull/51).

### DIFF
--- a/conf/iCubFindDependencies.cmake
+++ b/conf/iCubFindDependencies.cmake
@@ -72,8 +72,8 @@ checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
   # icub-firmware-shared 4.0.7 was actually a wrong version exported by icub-firmware-shared <= 1.15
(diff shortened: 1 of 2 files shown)


pattacini (approved):
Waiting CI before merging.

PR #780, opened 2022-01-17
added default values for FT58 and FT45

triccyx:
@marcoaccame 
I have added two new params:
- j4 set the gains to FT45 default
- j5 set the gains to FT58 default

-j is still present and is mapped on -j4

Could you verify if this approach is good for you? Otherwise, we can discuss an alternative.

@davidetome @pattacini

### DIFF
--- a/src/tools/FirmwareUpdater/main.cpp
+++ b/src/tools/FirmwareUpdater/main.cpp
@@ -5,8 +5,6 @@
 #include <QFileInfo>
 #include <qdebug.h>
 #include <QDir>
-#include <iostream>//LUCA
-
 
 #include "firmwareupdatercore.h"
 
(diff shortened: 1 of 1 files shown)


pattacini:
Hi @triccyx 

What's the rationale for

>- j4 set the gains to FT45 default
>- j5 set the gains to FT58 default

❔

Why don't we go for `j45` and `j58` instead?

triccyx:
It is a mere technical problem.
I have reused the `j` parameters and created a couple of other 2 parameters `4` and `5`. So used together become `-j4` and `-j5`.
It is not possible to create a parameter with more than 1 letter/digit.

triccyx:
But it occurred to me now that I can also give a value to the `j` param like this
`-j FT45`
perhaps is better

pattacini:
>perhaps is better

Definitely better, yeah 👍🏻

triccyx:
> > perhaps is better
> 
> Definitely better, yeah 👍🏻

Ok, I check how to change the code.

triccyx:
All the available letters for parameters are already used (a...z) also some numbers! In order to maintain retro compatibility for the old `-j` parameters I choose the following solution:
`-3 FT45` set the gains to FT45 default
`-3 FT58` set the gains to FT58 default
`-j` set the gains to FT45 default for retro compatibility
Unfortunately with the current library, used for parameters reading, seems that is not possible to specify a parameter (`-j`) with and without a value another way the best things would be:
`-j FT45` set the gains to FT45 default
`-j FT58` set the gains to FT58 default
`-j` set the gains to FT45 default

davidetome (approved):

davidetome (approved):

pattacini (approved):

PR #804, opened 2022-06-07
Fix net performance tool in embObjLib

valegagge:
Alongside with @vvasco, we fixed the tools inside the embObjLib for test network performance.

In details:
- we updated the cmake file: now the option `NETWORK_PERFORMANCE_BENCHMARK`, that enables the tool, works properly.
- we fixed `Emb_PeriodicEventVerifier`: now it uses the correct measure units.
- we fixed a bug about the casting from int to double.
- we updated the print histogram function.

We already used these changes to test the network performance on a robot. All is fine.


pattacini (approved):
Thanks @valegagge @vvasco 👍🏻 
Fine with me.
Don't know whether @marcoaccame wants to review it otherwise I'll be happy to merge the PR.

marcoaccame (approved):
seems fine to me.

PR #813, opened 2022-07-01
Added canBusFtSensor with ISixAxisForceTorqueSensors interface

vvasco:
This PR introduces `canBusFtSensor` which is a copy of `canBusAnalogSensor` with the implementation of the `ISixAxisForceTorqueSensors` interface.
It also adds two additional parameters as they are exposed by the interface, as suggested by @randaz81:
- `sensorName` mandatory
- `frameName` optional as it is used by ROS (default value is `unknown_frame_name`)

```xml
    <device xmlns:xi="http://www.w3.org/2001/XInclude" type="canBusFtSensor" name="ftSense">
        <param name="canbusDevice"> "sharedcan" </param>
        <param name="physDevice"> "socketcan" </param>
        <param name="canDeviceNum"> "can0" </param>
        <param name="canAddress"> 0x0D </param>
        <param name="format"> 16 </param>
        <param name="channels"> 6 </param>
        <param name="period"> 1 </param>
        <param name="useCalibration"> 1 </param>
        <param name="diagnostic"> 1 </param>
        <param name="sensorName"> strain </param>
        <param name="frameName"> id_l_arm_strain </param>
    </device>
```
I didn't edit directly `canBusAnalogSensor` to avoid retrocompatibility issue (`sensorName` mandatory). Eventually `canBusAnalogSensor` could be deprecated in favor of `canBusFtSensor` and the related robot xml updated.


pattacini (commented):
>I didn't edit directly canBusAnalogSensor to avoid retrocompatibility issue (sensorName mandatory). Eventually canBusAnalogSensor could be deprecated in favor of canBusFtSensor and the related robot xml updated.

I can see that `canBusAnalogSensor` is also used for the MAIS (e.g., https://github.com/robotology/robots-configuration/search?p=1&q=canBusAnalogSensor). Hence, it seems that there's no one-to-one correspondence between the old and the new service, at least for the name.

Can you clarify?

vvasco:
> > I didn't edit directly canBusAnalogSensor to avoid retrocompatibility issue (sensorName mandatory). Eventually canBusAnalogSensor could be deprecated in favor of canBusFtSensor and the related robot xml updated.
> 
> I can see that `canBusAnalogSensor` is also used for the MAIS (e.g., [robotology/robots-configuration/search?p=1 (canBusAnalogSensor)](https://github.com/robotology/robots-configuration/search?p=1&q=canBusAnalogSensor)). Hence, it seems that there's no one-to-one correspondence between the old and the new service, at least for the name.
> 
> Can you clarify?

Hi @pattacini, `canBusFtSensor` exposes the interface [`ISixAxisForceTorqueSensors`](https://www.yarp.it/git-master/classyarp_1_1dev_1_1ISixAxisForceTorqueSensors.html) that is used only for FT sensors. For the MAIS and all the other analog sensors, we could still rely on `canBusAnalogSensor`.

pattacini:
>Hi @pattacini, canBusFtSensor exposes the interface [ISixAxisForceTorqueSensors](https://www.yarp.it/git-master/classyarp_1_1dev_1_1ISixAxisForceTorqueSensors.html) that is used only for FT sensors. For the MAIS and all the other analog sensors, we could still rely on canBusAnalogSensor.

Ok, so we won't deprecate `canBusAnalogSensor` in the long term.

pattacini (approved):

PR #816, opened 2022-08-04
fix voltage truncation and add some doc

triccyx:
Fix voltage truncation to the first decimal.
Fix documentation about debugging.


pattacini (approved):
Thanks @triccyx for the fix!

PR #821, opened 2022-09-09
Fix frame name for multipleFT sensors

triccyx:
Fix frame name for multipleFTSensors
This is the first step done for MultipleFtSensor to understand if this fix is correct @randaz81


pattacini (commented):
Hi @triccyx 

As discussed, let's extend the fix to the single FT device too.

pattacini:
Just giving @randaz81 a couple of days for reviewing the PR.
I'm gonna merge it this week.

PR #823, opened 2022-09-14
Update `eomcParser` to set a default value (zero) for the friction parameters when they are not included in the config files.

sgiraz:
What's new in this PR:
- Update `eomcParser` to set a default value (zero) for the friction parameters when they are not included in the config files.

**Notes:**
- When friction parameters are not included in configuration files, the `yarprobotinterface` log shows a warning instead of an error and sets the friction parameters to zero. 
- Tested on AEA2 setup

cc @pattacini @isorrentino @marcoaccame


pattacini (approved):
Thanks @sgiraz 🚀 
Waiting for the CI before merging.

PR #827, opened 2022-09-16
Update EthMantainer to support discovering in unicast mode

sgiraz:
What's new:
- Add a new macro `ETH_MAINTAINER_DISCOVER_UNICASTMODE` to enable the `FirmwareUpdater` to discover:
  - The default address for boards like `AMC` and `AMCBLDC` --> `10.0.1.99`
  - Any address between `10.0.1.1` and `10.0.1.32` assigned by the user
  
**Notes:**
None


marcoaccame (approved):
The change is OK.

PR #829, opened 2022-09-23
Add support for AksIM2

sgiraz:
What's new:

- [Work in progress]

**Notes:**

### DIFF
--- a/conf/iCubFindDependencies.cmake
+++ b/conf/iCubFindDependencies.cmake
@@ -72,8 +72,8 @@ checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
   # icub-firmware-shared 4.0.7 was actually a wrong version exported by icub-firmware-shared <= 1.15
-  if(icub_firmware_shared_VERSION VERSION_GREATER 4 OR icub_firmware_shared_VERSION VERSION_LESS 1.27.0)
-    message(FATAL_ERROR "An old version of icub-firmware-shared has been detected, but at least 1.27.0 is required")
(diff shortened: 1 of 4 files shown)


conf/iCubFindDependencies.cmake
```diff
@@ -72,8 +72,8 @@ checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
   # icub-firmware-shared 4.0.7 was actually a wrong version exported by icub-firmware-shared <= 1.15
-  if(icub_firmware_shared_VERSION VERSION_GREATER 4 OR icub_firmware_shared_VERSION VERSION_LESS 1.27.0)
-    message(FATAL_ERROR "An old version of icub-firmware-shared has been detected, but at least 1.27.0 is required")
+  if(icub_firmware_shared_VERSION VERSION_GREATER 4 OR icub_firmware_shared_VERSION VERSION_LESS 1.27.1)
```
pattacini, line 75: This is a general remark, and as such don't take it as a blocker.

Since you're adding a new feature, which I presume doesn't break any back-compatibility, you should increase the minor version and not the patch number. We adhere to [semantic versioning](https://semver.org/), in fact.

Of course, the `icub-firmware-shared` suite goes along with the firmware version, which is a sort of constraint. At any rate, I think it's a good point to discuss.

There's no fixes, only new features, hence we should go w/ 1.28.0 and not 1.27.1.

cc @marcoaccame

sgiraz: Hi @pattacini, 

Thank you for your advise, I'll take this into account for the next PRs.

pattacini: Oki 👍🏻 
Waiting for the CI and then happy to merge your contribution!

pattacini (approved):

PR #830, opened 2022-09-27
Fix for Unittest on Windows

triccyx:
Fix for Unittest on Windows


pattacini (approved):
Super! Thanks @triccyx!

PR #839, opened 2022-10-24
Fix handling of limbs versions in `iKin`

pattacini:
This PR follows up on comments reported in #838 for which limbs version 2.10 is handled equally to 2.1 in iKin, for example.

To remedy this, a dedicated class `iKinLimbVersion` is introduced and thus all the related components have been updated.

This is an old flaw that emerges now as we have crafted the new iCub version 2.10.

The gaze and cartesian controller client/server versions have been upgraded to 2.0 to notify a breaking change in the way `getInfo()` operates for the hardware-version[^1].

Careful tests need to be done.

[^1]: Other code that needs to be upgraded:
      - https://github.com/robotology/stereo-vision/blob/master/lib/src/disparityThread.cpp#L381
      - https://github.com/robotology/icub-basic-demos/blob/master/demoRedBall/src/main.cpp#L706

### DIFF
--- a/src/modules/iKinGazeCtrl/src/main.cpp
+++ b/src/modules/iKinGazeCtrl/src/main.cpp
@@ -578,7 +578,7 @@ Windows, Linux
 #include <iCub/solver.h>
 #include <iCub/controller.h>
 
-#define GAZECTRL_SERVER_VER     1.2
+#define GAZECTRL_SERVER_VER     "2.0"
 
 using namespace std;
 using namespace yarp::os;
@@ -765,11 +765,11 @@ protected:
 
         Bottle &serverVer=info.addList();
         serverVer.addString("server_version");
-        serverVer.addFloat64(GAZECTRL_SERVER_VER);
+        serverVer.addString(GAZECTRL_SERVER_VER);
 
         Bottle &headVer=info.addList();
         headVer.addString("head_version");
-        headVer.addFloat64(commData.head_version);
+        headVer.addString(commData.head_version.get_version());
 
         Bottle &minVer=info.addList();
         minVer.addString("min_allowed_vergence");
@@ -1067,25 +1067,25 @@ protected:
     }
 
     /************************************************************************/
-    double constrainHeadVersion(const double ver_in)
+    iKinLimbVersion constrainHeadVersion(const iKinLimbVersion &ver_in)
     {
         // std::map<k,v> is ordered based on std::less<k>
-        map<double,double> d;
-        d[fabs(1.0-ver_in)]=1.0;
-        d[fabs(2.0-ver_in)]=2.0;
-        d[fabs(2.5-ver_in)]=2.5;
-        d[fabs(2.6-ver_in)]=2.6;
-        d[fabs(2.7-ver_in)]=2.7;
-        d[fabs(2.8-ver_in)]=2.8;
-        d[fabs(2.10-ver_in)]=2.10;
-        d[fabs(3.0-ver_in)]=3.0;
-        d[fabs(3.1-ver_in)]=3.1;
+        map<iKinLimbVersion,iKinLimbVersion> d;
+        d[iKinLimbVersion("1.0")-ver_in]=iKinLimbVersion("1.0");
+        d[iKinLimbVersion("2.0")-ver_in]=iKinLimbVersion("2.0");
+        d[iKinLimbVersion("2.5")-ver_in]=iKinLimbVersion("2.5");
+        d[iKinLimbVersion("2.6")-ver_in]=iKinLimbVersion("2.6");
+        d[iKinLimbVersion("2.7")-ver_in]=iKinLimbVersion("2.7");
+        d[iKinLimbVersion("2.8")-ver_in]=iKinLimbVersion("2.8");
+        d[iKinLimbVersion("2.10")-ver_in]=iKinLimbVersion("2.10");
+        d[iKinLimbVersion("3.0")-ver_in]=iKinLimbVersion("3.0");
+        d[iKinLimbVersion("3.1")-ver_in]=iKinLimbVersion("3.1");
 
-        double ver_out=d.begin()->second;
+        auto ver_out=d.begin()->second;
         if (ver_out!=ver_in)
         {
-            yWarning("Unknown \"head_version\" %g requested => used \"head_version\" %g instead",
-                     ver_in,ver_out);
+            yWarning("Unknown \"head_version\" %s requested => used \"head_version\" %s instead",
+                     ver_in.get_version().c_str(),ver_out.get_version().c_str());
         }
 
         return ver_out;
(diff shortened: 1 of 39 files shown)


src/modules/iKinGazeCtrl/src/main.cpp
```diff
@@ -1067,25 +1067,25 @@ class GazeModule: public RFModule
...
+        map<iKinLimbVersion,iKinLimbVersion> d;
+        d[iKinLimbVersion("1.0")-ver_in]=iKinLimbVersion("1.0");
+        d[iKinLimbVersion("2.0")-ver_in]=iKinLimbVersion("2.0");
+        d[iKinLimbVersion("2.5")-ver_in]=iKinLimbVersion("2.5");
+        d[iKinLimbVersion("2.6")-ver_in]=iKinLimbVersion("2.6");
+        d[iKinLimbVersion("2.7")-ver_in]=iKinLimbVersion("2.7");
+        d[iKinLimbVersion("2.8")-ver_in]=iKinLimbVersion("2.8");
+        d[iKinLimbVersion("2.10")-ver_in]=iKinLimbVersion("2.10");
+        d[iKinLimbVersion("3.0")-ver_in]=iKinLimbVersion("3.0");
+        d[iKinLimbVersion("3.1")-ver_in]=iKinLimbVersion("3.1");
```
pattacini, line 1082: This needs careful verification.

src/modules/iKinGazeCtrl/src/main.cpp
```diff
@@ -1142,7 +1142,7 @@ class GazeModule: public RFModule
         commData.robotName=rf.check("robot",Value("icub")).asString();
         commData.eyeTiltLim[0]=eyeTiltGroup.check("min",Value(-20.0)).asFloat64();
         commData.eyeTiltLim[1]=eyeTiltGroup.check("max",Value(15.0)).asFloat64();
-        commData.head_version=constrainHeadVersion(rf.check("head_version",Value(1.0)).asFloat64());
+        commData.head_version=constrainHeadVersion(iKinLimbVersion(rf.check("head_version",Value("1.0")).asString()));
```
pattacini, line 1145: This is potentially breaking.
However, the input param `1.0` from the command line should be interpreted as a string without problems.

Nicogene (approved):
At first glance seems ok 👍🏻

pattacini:
Before doing specific tests on the robot, I'll be verifying this solution in https://github.com/pattacini/icub-gazebo-grasping-sandbox.

src/modules/iKinGazeCtrl/src/main.cpp
pattacini: This snippet wasn't spot on.
Fully refactored. Currently under test.

src/modules/iKinGazeCtrl/src/main.cpp
pattacini: Unfortunately, `--head_version 1.0` or `--head_version "1.0"` seem not to work 😔
I need more testing. A possibility is to rely on `--head_version v1.0` but is a breaking change, as pretty much as "1.0".

pattacini: The following simple snippet demonstrates the use case:
```c++
#include <cstdlib>
#include <yarp/os/ResourceFinder.h>
#include <yarp/os/LogStream.h>

int main(int argc, char* argv[]) {
    yarp::os::ResourceFinder rf;
    rf.configure(argc, argv);
    auto option = rf.check("option", yarp::os::Value("1.0")).asString();
    yInfo() << "option:" << option;
    return EXIT_SUCCESS;
}
```

Outputs:
```console
$ test_2
[INFO] option: 1.0
$ test_2 --option 2.0
[INFO] option: 
$ test_2 --option "2.0"
[INFO] option: 
$ test_2 --option "v2.0"
[INFO] option: v2.0
```

pattacini: My understanding is that if I enforce `.asString()`, everything passed in ought to be interpreted as a string, actually, even in the format w/o quotes.

cc @Nicogene

pattacini: >My understanding is that if I enforce .asString(), everything passed in ought to be interpreted as a string, actually, even in the format w/o quotes.

Probably not because `.asString()` is only a way to _re-interpret_ the content of the Value, which was already filled in.

@Nicogene do you know if there exists a simple way to get around it?

Nicogene: What about getting the version with `rf.check("head_version",Value(1.0)).asFloat64())` as before, and then convert it to string with [`std::to_string`](https://en.cppreference.com/w/cpp/string/basic_string/to_string) for passing it to `iKinLimbVersion`?

Nicogene: I would also add this version check tests in the icub-main unit tests

pattacini: This won't work as 2.1==2.10 for floats, neglecting the precision.

I thought about extending the tests but it won't be worthwhile covering an extremely little part of iKin.

The change will be breaking, most likely. No problem, I'll handle that.

mfussi66 (approved):

pattacini:
> Before doing specific tests on the robot, I'll be verifying this solution in https://github.com/pattacini/icub-gazebo-grasping-sandbox.

After various fixes, I managed to make it work. See the result below:

https://user-images.githubusercontent.com/3738070/201530414-7fdbc2f8-edbc-4707-ac7c-9f865fcf17d0.mp4

For the record, here's enclosed the patch file to adapt the sandbox: [`master...pattacini_icub-gazebo-grasping-sandbox_master.patch.txt`](https://github.com/robotology/icub-main/files/9997630/master.pattacini_icub-gazebo-grasping-sandbox_master.patch.txt).

Verifying that the sandbox works as intended is not sufficient:
- The next move is to assess the PR on a physical robot.
- Then, we're required to update the configuration file for the iKinGazeCtrl module.

pattacini:
>- Then, we're required to update the configuration files for the iKinGazeCtrl module.

See:
- https://github.com/robotology/robots-configuration/pull/397

xEnVrE:
In [realsense-holder-calibration](https://github.com/robotology/realsense-holder-calibration/blob/fb94c261c6d44150e056671a5f482235e7ecc540/src/forward-kinematics/src/ForwardKinematics.cpp#L28), that is linked in https://icub-tech-iit.github.io/documentation/upgrade_kits/realsense_holder/support/#assembly-instruction, we also have a snippet of code of the form

```
icub_head_center_ = iCub::iKin::iCubHeadCenter("right_" + head_version);
```

Does that need the same kind of patch as done in `stereo-vision`?

pattacini:
>Does that need the same kind of patch as in stereo-vision?

At first sight, I think it's ok as that module employs the conf param `eye_version`, which contains the "v" char and there doesn't seem to be any version comparison.
See:
- https://github.com/search?q=repo%3Arobotology%2Frealsense-holder-calibration%20eye_version&type=code
- https://github.com/robotology/realsense-holder-calibration/blob/fb94c261c6d44150e056671a5f482235e7ecc540/src/calibration/app/conf/config_iCubGenova01_holder_no_tilt.ini#L4

At any rate, better off an deeper inspection.

Nicogene (changes requested):
SInce it is a breaking change I would increment the patch number, in this way we could set the new icub-main version as minimum required for the downstream code that uses iKin.

pattacini:
> SInce it is a breaking change I would increment the patch number, in this way we could set the new icub-main version as minimum required for the downstream code that uses iKin.

Actually, I was thinking to increment the major number, right because it's a breaking change.

pattacini:
>Actually, I was thinking to increment the major number, right because it's a breaking change.

Done, @Nicogene!

Nicogene (approved):

pattacini:
> > * Then, we're required to update the configuration files for the iKinGazeCtrl module plus the other 2 downstream projects.
> 
> See:
> 
> * [[iKinGazeCtrl] patch versioning robots-configuration#397](https://github.com/robotology/robots-configuration/pull/397)
> * [Patch iKin versioning stereo-vision#36](https://github.com/robotology/stereo-vision/pull/36)
> * [Patch iKin versioning icub-basic-demos#38](https://github.com/robotology/icub-basic-demos/pull/38)

I've merged the PR's above too.

cc @Nicogene

PR #843, opened 2022-11-25
Build YCM and YARP required versions for CI

pattacini:
This PR proposes to build the CI by cloning the exact required versions of YCM and YARP.


pattacini:
`icub-main` is built against `ycm@master` and `yarp@master` anyway relying on the superbuild infrastructure.

PR #846, opened 2022-12-05
Fix boolean expression when checking type of strain2 board

mfussi66:
This PR fixes the boolean expression that is used when checking the type of strain2 board.

If it's `OR` and the board is strain2, the `yarprobotinterface` fails the launch. Otherwise, with `AND` it fails only when the board is different from strain2 and strain2c


mfussi66:
cc @AntonioConsilvio @pattacini @marcoaccame @davidetome

pattacini (approved):

pattacini:
It makes sense 👍🏻 
Merging it straight away to go quicker and rebasing #839[^1].

[^1]: We can always amend it later on.

PR #847, opened 2022-12-06
[iKin] fixed clash with system macros

pattacini:
Apparently, `major` and `minor` are system-defined macros on some platforms.

See https://github.com/robotology/robotology-superbuild/issues/1313.


traversaro (approved):
Thanks!

PR #858, opened 2023-02-24
POS can use also `mtb4` and `mtb4c` boards

marcoaccame:
This PR allows the correct parsing of the POS section of the xml files used for MC in the new hand of `ergoCub` with also the `mtb4` and `mtb4c` boards.
It is associated to:
- https://github.com/robotology/icub-firmware/pull/347
- https://github.com/robotology/icub-firmware-shared/pull/79
- https://github.com/robotology/icub-firmware-build/pull/69


pattacini (approved):

PR #861, opened 2023-04-05
Delete grabberDual references

martinaxgloria:
As discussed in [#469](https://github.com/robotology/robots-configuration/pull/469), this PR aims to delete the `grabberDual` references already present in icub-main.

### DIFF
--- a/app/default/scripts/cameras_calib.xml.template
+++ b/app/default/scripts/cameras_calib.xml.template
@@ -4,12 +4,12 @@
 </dependencies> 
 <module>
   <name>yarpdev</name>
-      <parameters> --from camera/dragonfly2_config_left.ini --split true</parameters>
+      <parameters> --from camera/dragonfly2_config_left.ini </parameters>
   <node>pc104</node>
 </module>
 <module>
   <name>yarpdev</name>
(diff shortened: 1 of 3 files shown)


app/default/scripts/cameras_calib.xml.template
```diff
@@ -4,7 +4,12 @@
 </dependencies> 
 <module>
   <name>yarpdev</name>
-      <parameters> --from camera/ServerGrabberDualDragon.ini --split true</parameters>
+      <parameters> --from camera/dragonfly2_config_left.ini --split true</parameters>
```
pattacini, line 7: The parameter `--split true` should be only meaningful with the dual driver.
@Nicogene, could you confirm this?
If confirmed, we ought to remove it.

pattacini (changes requested):
Thanks @martinaxgloria 👍🏻 
Could you:
- Sign the CLA?
- Address the inline comment?

app/default/scripts/cameras_calib.xml.template
Nicogene: Yes I confirm it is obsolete 👍🏻

Nicogene (approved):

pattacini (approved):

PR #869, opened 2023-05-09
Added the mrie encoder name in the parser of motion control service.

valegagge:
Now we can use the `mrie` name in the configuration file.
This encoder is used for the new joints.

I already tested this PR on the medium joint setup and it works together.

This code needs the updated `icub-firmware-shared` at version `1.34.2`.
I'll open a new PR for it.

### DIFF
--- a/conf/iCubFindDependencies.cmake
+++ b/conf/iCubFindDependencies.cmake
@@ -65,7 +65,7 @@ checkandset_dependency(OpenCV)
 checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
-  if(icub_firmware_shared_VERSION VERSION_LESS 1.34.1)
+  if(icub_firmware_shared_VERSION VERSION_LESS 1.34.2)
     message(FATAL_ERROR "An old version of icub-firmware-shared has been detected: at least 1.34.1 is required")
   endif()
 endif()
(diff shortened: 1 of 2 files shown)


conf/iCubFindDependencies.cmake
```diff
@@ -65,7 +65,7 @@ checkandset_dependency(OpenCV)
 checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
-  if(icub_firmware_shared_VERSION VERSION_LESS 1.34.1)
+  if(icub_firmware_shared_VERSION VERSION_LESS 1.34.2)
     message(FATAL_ERROR "An old version of icub-firmware-shared has been detected: at least 1.34.1 is required")
```
pattacini, line 69: >This code needs the updated icub-firmware-shared at version 1.34.3.

This doesn't match...
Any comment on this?

pattacini: Also, we'd need to update the error message.

pattacini (changes requested):
Put the PR in draft[^1] as the CI will fail most likely since we need to be aligned with `icub-firmware-shared`. 

[^1] For a draft PR we don't run CI.

conf/iCubFindDependencies.cmake
pattacini, line 69:
```suggestion
    message(FATAL_ERROR "An old version of icub-firmware-shared has been detected: at least 1.34.2 is required")
```

pattacini (approved):
Turning on the CI

PR #874, opened 2023-05-19
embObjFTSensor: make it handle strain board

### LINKED ISSUE #873: Make `embObjFTSensor` handle strain board for removing `embObjStrain` device

Right now `embObjFTSensor` device cannot handle the `strain` board because it expect to have also a temperature sensor that `strain` does not have. It is giving problems for the refactoring on how we expose FT data, see https://github.com/robotology/robots-configuration/issues/485#issuecomment-1552844722.

Make `embObjFTSensor` handle strain board, the nr of temperature sensor will be set to 0 and `temperature-aquisitionRate` will be ignored at the startup. Moreover, we can get rid of `embObjStrain` 

cc @traversaro @pattacini 

_No response_

### Additional context

_No response_

Nicogene:
It fixes #873.

I will test it on the robot before merging.

### DIFF
--- a/src/libraries/icubmod/embObjFTsensor/embObjFTsensor.cpp
+++ b/src/libraries/icubmod/embObjFTsensor/embObjFTsensor.cpp
@@ -95,7 +95,7 @@ bool embObjFTsensor::open(yarp::os::Searchable &config)
         return false;
 
     // read stuff from config file
-    
+
     servConfigFTsensor_t serviceConfig;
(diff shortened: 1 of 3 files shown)


Nicogene:
Successfully tested on `iCubGenova02`, I can read now both from strain and strain2 using MASserver

cc @traversaro

pattacini (approved):
Nice!

PR #877, opened 2023-05-24
Change the device running faceExpressions from serialport to serialPort_nws_yarp

martinaxgloria:
With the last releases, the old `serialport` device used for running the faceExpressions.xml application could not be found (see figure below) 

MicrosoftTeams-image (7)

In this way, it is replaced with the new `serialPort_nws_yarp` which runs properly on iCubGenova11.


pattacini (approved):
Thanks heaps!

PR #883, opened 2023-05-30
Fix serialPort_nws_yarp device to run faceExpressions on iCub*

martinaxgloria:
I removed the `faceExpressionsSim.xml.template` since it's no more used and I added a new file called `serialPortFaceExpressions.xml`. The reason is that the device `serialPort_nws_yarp` could not be run with `yarpdev` which is deprecated (cc @randaz81) but it should be launched with the `yarprobotinterface` using the new xml file as `config` parameter.

This new configuration was tested on iCubGenova11 and it works properly, also in changing the face expressions through the rpc protocol.


randaz81:
For extra clarity: yarpdev is not currently deprecated, but it is unmaintained and it might be deprecated in the near future. Some old devices will still work, but new nws devices are developed and tested to be used only with yarprobotinterface.
Besides having a different plugin loader, yarprobotinterface supports attach operations between multiple devices, which is a fundamental feature for ROS/ROS2 compatibility.

pattacini (approved):

PR #885, opened 2023-06-05
Calibration type 14 update on calibration value parser

MSECode:
This PR brings an update to the current version of `calibration type 14`. Specifically the following things have changed:
- FAP calibration data are not passed anymore only through the `CALIBRATION` group in the `POS` service configuration file but the final correct values should be defined at `parameters 3, 4 and 5 ` of the calibration configuration file as defined in the documentation. Moreover in this file `offset` and `rotation` values should be passed in `icubDegrees`. 
- Anyway, `CALIBRATION` group in MC and POS is still kept for debugging reasons. Values in calibrators will override the one configured initially by the POS service.
- Then, with this update the calibration will do the following steps:
    - calibrate the FAP absolute encoder
    - bring the motor pulley to the hard stop
- New data structure have been introduced to simplify the conversion between eoThePOS datatype and icubDegrees to be used by both `icub-main` and `icub-firmware`
- Add check for `calib.params.type14.invertdirection` value, which must be a boolean value
- Further specification can be found in the [documentation ](https://icub-tech-iit.github.io/documentation/icub_r1_icub3_calibration_types/icub_r1_calibration_types/?h=calibrations#the-available-calibrations-in-details)

### DIFF
--- a/conf/iCubFindDependencies.cmake
+++ b/conf/iCubFindDependencies.cmake
@@ -65,8 +65,8 @@ checkandset_dependency(OpenCV)
 checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
-  if(icub_firmware_shared_VERSION VERSION_LESS 1.34.2)
-    message(FATAL_ERROR "An old version of icub-firmware-shared has been detected: at least 1.34.2 is required")
+  if(icub_firmware_shared_VERSION VERSION_LESS 1.34.3)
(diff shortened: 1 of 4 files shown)


pattacini:
Also, please fill in the body of the PR.

marcoaccame (approved):
The PR is OK @MSECode, thanks.

pattacini:
We're still missing the version check for `icub-firmware-shared`.

pattacini (approved):
Thanks @MSECode 👍🏻 
Once the upstream PR is merged, I'll put this in ready for review to activate the CI.

Issue #880, opened 2023-05-26
Wrong position of triangles 19, 20 and 21 in right_leg_lower.ini


BoPot:
### Bug description

The positions and orientations of triangles 19, 20 and 21 given in the configuration file to lower right leg skin are different from reality.
@matejhof @rustlluk @traversaro 

### Steps to reproduce

My code for reading the configuration file and plotting the skin model:
https://gitlab.fel.cvut.cz/body-schema/icub/icub-skin-calibration/-/blob/main/skin_models/read_ini.py

### Additional context

The model generated from right_leg_lower.ini:
![right_flat_19_20_21_marked](https://github.com/robotology/icub-main/assets/8711335/fc272d21-825d-418e-a4ab-c210121c33e6)
Reality:
![right_reality_19_20_21_marked](https://github.com/robotology/icub-main/assets/8711335/4e0acb97-968c-4015-b882-3496d002f157)

pattacini:
Thanks for reporting @BoPot 👍🏻 
We will get back to this in two weeks as we will be attending ICRA.

/remind June 07

PR #889, opened 2023-06-20
fixed triangle positions as per #880

simeonedussoni:
the position of triangles for right_leg_lower is fixed, starting from the `ini_generator` excel spreadsheet up to the `.ini` file relative to that part.


pattacini (approved):
Thanks @simeonedussoni 👍🏻

PR #876, opened 2023-05-19
wbd: port it for using MAS client also for FT 

Nicogene:
It fixes #865 

x-ref: #596 

TO BE TESTED!

PR #890, opened 2023-06-21
wbd: fix dummy_ft options after refactoring

Nicogene:
This regression has been added in #876


pattacini (approved):
Tested the patch w/ Gazebo and it worked like a charm 🚀

PR #866, opened 2023-04-27
EthCalibrator: fixed CheckHwFault

valegagge:
During the investigation of [this ticket](https://github.com/robotology/icub-tech-support/issues/1544), @MSECode and I noticed some bugs in the `Ethernet calibrator`.

 1. there is an empty debug string: `DEBUG] Left_Arm_Calibrator : Joints calibration order:      ` 
2. the `checkHWFault` function performed a loop on the joints in (0, totJointsToCalibrate) range instead of on the list containing the required joints.
3. the return value of `checkHWFault` function was ignored.

We fixed them in this way:
1. 
Screenshot from 2023-04-27 12-03-54

2. the loop in performed on the elements containing in the list of joints to calibrate 

| Before | After |
|---|---|
| Screenshot from 2023-04-27 12-06-07 | Screenshot from 2023-04-27 12-06-50 |

3. Added a check on the return value of `checkHWFault` function
Screenshot from 2023-04-27 12-09-23

## Results:
Here there are the logs of the left arm during calibration where the joint number 9 is omitted

**Before**
``` 
[DEBUG] left_arm-calibrator starting calibration of device left_arm-mc_remapper
[INFO] Left_Arm_Calibrator : starting calibration
[ERROR] Left_Arm_Calibrator cannot perform 'check hardware fault' operation because joint number 9 is out of range [].
[DEBUG] Left_Arm_Calibrator : Joints calibration order:
[WARNING] Left_Arm_Calibrator  is calibrating only a subset of the robot part. 

```

**After**
```

[DEBUG] left_arm-calibrator starting calibration of device left_arm-mc_remapper
[INFO] Left_Arm_Calibrator : starting calibration
[DEBUG] Left_Arm_Calibrator : Joints calibration order: 4 5 6 3 2 0 1 7 8 10 11 12
[WARNING] Left_Arm_Calibrator  is calibrating only a subset of the robot part. Calibrating  12  over a total of  13

```

PR #898, opened 2023-08-09
Temporary workaround for MAIS finger problem

MSECode:
This PR introduces a temporary fix to the hw fault check in order to do not have problem with YRI restarting on joints that mounts encoders which are highly sensible to disturbances, such as the MAIS on the fingers of iCub.
The reason to introduce this change and to define it temporary are here defined:
- previously with [this PR](https://github.com/robotology/icub-main/pull/866) the check on the hw faults has been enabled. As a matter of fact we realized that in that code section mentioned in the PR the logic of the code was erroneous, because the check of the faults is never taken into account, but the operations can continue independently on the fact that a fault is present or not. Therefore, it has been decided to wrap the method in a if condition and stop the calibration procedure if a fault is present;
- however, that PR brought up some misbehavior of the code, which were previously hidden, such as the fact that if an encoder is sensible to errors, it could happen that it measure oscillates around zero at YRI start-up and if it goes towards negative values  there will be an hw limits error, which won't allow the code to go on with the calibration
- anyway, it should be underlined that the hw limit/boundaries exceed error happens because at YRI stopping the joints are set to IDLE instead of to NOT CONFIGURED, as when the YRI is started the first time. However, considering that modifying and test all the code of the motion controller for setting the joints to NOT CONFIGURED at YRI stop is planned for the upcoming release, as things now this fix is necessary to allow the users to work fine with MAIS joints.
- For sure this is not the best option but, after recalibrating the MAIS joints multiple times and having every time a different finger joint to go on fault, introduce the necessity to operates in this way

### DIFF
--- a/src/libraries/icubmod/parametricCalibratorEth/parametricCalibratorEth.cpp
+++ b/src/libraries/icubmod/parametricCalibratorEth/parametricCalibratorEth.cpp
@@ -578,8 +578,13 @@ bool parametricCalibratorEth::calibrate()
     }
 
     //before starting the calibration, checks for joints in hardware fault, and clears them if the user set the clearHwFaultBeforeCalibration option
-    if(!checkHwFault())
-        return false;
+    // if(!checkHwFault())
+    //     return false;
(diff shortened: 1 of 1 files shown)


MSECode:
cc:
@maggia80 @AntonioConsilvio @sgiraz

pattacini (approved):

PR #902, opened 2023-08-29
Update eomcParser and embObjMotionControl to support the new velocityThresh param

sgiraz:
**What's new:**
- The `eomcParser` now detects the presence of the new `velocityThresh` when it is written in the configuration files. If the `velocityThresh` is omitted its default value is 0.
- ⚠️ The `eomcParser` will not recognize the `kbemf` parameter. Hence, all the config files in robots-configuration have to be updated accordingly

> **Note**
> - This is connected to
>   - https://github.com/robotology/yarp/pull/3008
>   - https://github.com/robotology/icub-firmware-shared/pull/86

### DIFF
--- a/src/libraries/icubmod/canBusMotionControl/CanBusMotionControl.cpp
+++ b/src/libraries/icubmod/canBusMotionControl/CanBusMotionControl.cpp
@@ -1463,9 +1463,6 @@ bool CanBusMotionControlParameters::fromConfig(yarp::os::Searchable &p)
            }
            else
            {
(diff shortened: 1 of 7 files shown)


pattacini (changes requested):
Hi @sgiraz,

Given this [change](https://github.com/robotology/icub-firmware-shared/pull/86/files/f311458797e24ce46115a549114c9467191972db..69d2691f841d361a0d252cc4657de1d780566ca9), you ought to update the dependencies here too.

pattacini:
Hi @sgiraz 

Any reason why you opened another PR (see #903)?

PR #904, opened 2023-09-04
Correction of leg skin calibration

BoPot:
When creating the position files I uploaded earlier there was a minor error in my code, which caused the triangles to be oriented incorrectly (therefore wrong placement of taxels in individual triangles).
I have fixed the error and generated new position files. We have also tested the files on the robot with @matejhof and @rustlluk and everything seemed correct.


pattacini (approved):
Thank you @BoPot!

PR #906, opened 2023-09-07
Re-enabled the check of the hw fault before proceeding with calibration

MSECode:
Since, now the procedure for setting the control mode of the joints has been clarified, it is correct to re-enable the check of the hw faults before proceeding with the calibration procedure of the jonts.


pattacini (approved):

valegagge:
thanks @pattacini !

valegagge:
This PR works with this FW https://github.com/robotology/icub-firmware/pull/413 and build: https://github.com/robotology/icub-firmware-build/pull/102

PR #911, opened 2023-10-24
Feature/temperature reading

MSECode:
# Description:
This PR brings the following changes:
- user can pass to the lower level the temperature sensor type used by the motor related to the specific joint
- user can pass to the lower level the temperature limits used by the motor related to the specific joint

Here follows the details about the implemented changes and an action diagram describing the flow of the calls and the entities that takes part in the workflow:

MotorTemperature-Feature

## Parsing new configuration
1. we have introduced three new parameters as follows:
    - `hardwareTemperatureLimits` and `warningTemperatureLimits` in the `hardware/motorControl` section in the `LIMITS` group. The  `hardwareTemperatureLimits` is the maximum temperature at which the motor can work, above which the motor is damaged. Instead the `warningTemperatureLimits` describes a warning temperature defined by the user to advise to reduce the current fed to the specific motor suggesting not increasing the temperature anymore, going to reach the hardware limit with the consequent fault of the robot.  The `warningTemperatureLimit` must be at least 85% of the `hardwareTemperatureLimit` in order to have a secure gap before hitting the hardware temperature limit of the motor and thus to work in conditions that should be safe for not damaging the motors.
    - in the `hardware/mechanical` section in the `FOC` group we have introduced the `TemperatureSensorType` parameter, which is going to substitute the current `HasTempSensor` parameter, which is now still accepted if set to zero without adding the new group, but which will soon be fully deprecated. The possible values of `TemperatureSensorType` are `PT100`, `PT1000` and `NONE`, depending on the temperature sensor mounted. 

2. This feature is retro-compatible --> it is possible to use the robot without making any changes to the current configuration. Therefore, if we don't set the limits and if we leave the parameter `HasTempSensor` in the `FOC` group to `zero` everything works fine and the temperature reading on the FOC boards is not enabled.

### Testing of the new configuration
We tested all possible configurations (at least we hope to have covered all cases 😸 ) both in dry-run and actively on the bench setup and on the robot ergoCub SN001 testing the correct parsing of the new configuration. At [this comment](https://github.com/icub-tech-iit/five/issues/178#issuecomment-1785322184) it is possible to find the outcome of the tests done and here follows a brief summary of what the user should expect:
Perhaps it would be worth pointing out that, as said before, for retro-compatibility, if the user sets `HasTempSensor` to 0 then any temperature sensor is configured and the motor temperature will not read. Otherwise, if the user sets `HasTempSensor` to 1 then the parser returns an error related to an inconsistent configuration because the sensor type and limits parameters are missing.

## embObjMotionControl device
1. implemented `getTempeartures()` and `getTemperature()`: first of all, it checks if a temperature sensor type has been enabled for the specific motor, otherwise it won't be possible to receive a certain temperature. Then it checks if the raw value is different from a specific error value `-5000` that is used by the 2FOC boards to signal that it had I2C communication error. We chose -5000 because it is outside of the possible range of raw values. In the case of I2C communication errors, it returns false else converts from the raw value coming from the boards to a value in Celsius Degree and returns true.
2. implemented `getTemperatureSensorType()`: is quite strait-forward. It basically returns the type of sensor configured for each motor. 
3. when the update function is called by the correct callback function we check if :
   - the temperature exceeds the warning threshold. If yes it gives a warning message to the user
   - the temperature values as the I2C error value (-5000). If yes gives an error. 

In both cases, we use two watchdog objects to avoid flooding the log with the above-mentioned warning.

### DIFF
--- a/conf/iCubFindDependencies.cmake
+++ b/conf/iCubFindDependencies.cmake
@@ -65,8 +65,8 @@ checkandset_dependency(OpenCV)
 checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
-  if(icub_firmware_shared_VERSION VERSION_LESS 1.37.1)
-    message(FATAL_ERROR "An old version of icub-firmware-shared has been detected: at least 1.37.1 is required")
+  if(icub_firmware_shared_VERSION VERSION_LESS 1.37.2)
+    message(FATAL_ERROR "An old version of icub-firmware-shared has been detected: at least 1.37.2 is required")
   endif()
 endif()
 
--- a/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
@@ -100,6 +100,7 @@ bool embObjMotionControl::alloc(int nj)
     _encodersStamp = allocAndCheck<double>(nj);
     _gearbox_M2J = allocAndCheck<double>(nj);
     _gearbox_E2J = allocAndCheck<double>(nj);
+    _temperatureFactor_degcel2raw = allocAndCheck<double>(nj);
     _deadzone = allocAndCheck<double>(nj);
     _foc_based_info=allocAndCheck<eomc::focBasedSpecificInfo_t>(nj);
     _trj_pids= new eomc::PidInfo[nj];
@@ -127,6 +128,7 @@ bool embObjMotionControl::alloc(int nj)
     _rotorsLimits.resize(nj);
     _jointsLimits.resize(nj);
     _currentLimits.resize(nj);
+    _temperatureLimits.resize(nj);
     _jsets.resize(nj);
     _joint2set.resize(nj);
     _timeouts.resize(nj);
@@ -147,6 +149,7 @@ bool embObjMotionControl::dealloc()
     checkAndDestroy(_encodersStamp);
     checkAndDestroy(_gearbox_M2J);
     checkAndDestroy(_gearbox_E2J);
+    checkAndDestroy(_temperatureFactor_degcel2raw);
     checkAndDestroy(_deadzone);
     checkAndDestroy(_impedance_limits);
     checkAndDestroy(checking_motiondone);
@@ -208,6 +211,7 @@ embObjMotionControl::embObjMotionControl() :
     _rotorsLimits(0),
     _jointsLimits(0),
     _currentLimits(0),
+    _temperatureLimits(0),
     _jsets(0),
     _joint2set(0),
     _timeouts(0),
@@ -219,6 +223,7 @@ embObjMotionControl::embObjMotionControl() :
 {
     _gearbox_M2J  = 0;
     _gearbox_E2J  = 0;
+    _temperatureFactor_degcel2raw = 0;
     _deadzone     = 0;
     opened        = 0;
     _trj_pids     = NULL;
@@ -789,8 +794,6 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
 
 
     ///////// GENERAL MECHANICAL INFO
-
-
     {
         if(!_mcparser->parseAxisInfo(config, _axisMap, _axesInfo))
             return false;
@@ -814,7 +817,7 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
 
         if (!_mcparser->parseAmpsToSensor(config, measConvFactors.ampsToSensor))
             return false;
-
+        
         //VALE: i have to parse GeneralMecGroup after parsing jointsetcfg, because inside generalmec group there is useMotorSpeedFbk that needs jointset info.
 
         if(!_mcparser->parseGearboxValues(config, _gearbox_M2J, _gearbox_E2J))
@@ -962,6 +965,9 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
         if(!_mcparser->parseCurrentLimits(config, _currentLimits))
             return false;
 
+        if(!_mcparser->parseTemperatureLimits(config, _temperatureLimits))
+            return false;
+
         if(!_mcparser->parseJointsLimits(config, _jointsLimits))
             return false;
 
@@ -973,7 +979,7 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
     if(serviceConfig.ethservice.configuration.type == eomn_serv_MC_foc)
     {
         std::string groupName = (static_cast<eObrd_type_t>(serviceConfig.ethservice.configuration.data.mc.foc_based.type) == eobrd_foc) ? "2FOC" : "AMCBLDC";
-        if(!_mcparser->parseFocGroup(config, _foc_based_info, groupName))
+        if(!_mcparser->parseFocGroup(config, _foc_based_info, groupName, _temperatureFactor_degcel2raw))
             return false;
     }
 
@@ -1395,7 +1401,7 @@ bool embObjMotionControl::init()
         motor_cfg.rotEncTolerance = _motorEncs[logico].tolerance;
         motor_cfg.hasHallSensor = _foc_based_info[logico].hasHallSensor;
         motor_cfg.hasRotorEncoder = _foc_based_info[logico].hasRotorEncoder;
-        motor_cfg.hasTempSensor = _foc_based_info[logico].hasTempSensor;
+        motor_cfg.hasTempSensor = _foc_based_info[logico].hasTemperatureSensor;
         motor_cfg.hasRotorEncoderIndex = _foc_based_info[logico].hasRotorEncoderIndex;
         motor_cfg.hasSpeedEncoder = _foc_based_info[logico].hasSpeedEncoder;
         motor_cfg.verbose = _foc_based_info[logico].verbose;
@@ -1403,6 +1409,7 @@ bool embObjMotionControl::init()
         motor_cfg.rotorIndexOffset = _foc_based_info[logico].rotorIndexOffset;
         motor_cfg.rotorEncoderType = _motorEncs[logico].type;
         motor_cfg.pwmLimit =_rotorsLimits[logico].pwmMax;
+        motor_cfg.temperatureLimit = (eOmeas_temperature_t)(_temperatureLimits[logico].hardwareTemperatureLimit * _temperatureFactor_degcel2raw[logico]); //passing raw value not in degree
         motor_cfg.limitsofrotor.max = (eOmeas_position_t) S_32(_measureConverter->posA2E(_rotorsLimits[logico].posMax, fisico ));
         motor_cfg.limitsofrotor.min = (eOmeas_position_t) S_32(_measureConverter->posA2E(_rotorsLimits[logico].posMin, fisico ));
         
@@ -3270,7 +3277,15 @@ bool embObjMotionControl::getKinematicMJRaw(int j, double &rotres)
     return false;
 }
 
-bool embObjMotionControl::getHasTempSensorsRaw(int j, int& ret)
+bool embObjMotionControl::getTemperatureSensorTypeRaw(int j, int& ret)
+{
+    // refresh cached value when reading data from the EMS
+    ret = (int)_foc_based_info[j].temperatureSensorType;
+
+    return true;
+}
+
+bool embObjMotionControl::getHasTempSensorsRaw(int j, int& ret) 
 {
     eOprotID32_t protoid = eoprot_ID_get(eoprot_endpoint_motioncontrol, eoprot_entity_mc_motor, j, eoprot_tag_mc_motor_config);
     uint16_t size;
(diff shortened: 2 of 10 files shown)


pattacini:
Awaiting https://github.com/robotology/icub-main/pull/911 to be merged before putting this in ready for review.

pattacini:
Hi @MSECode 

Could you please realign the required version of `icub-firmware-shared`?

MSECode:
Just done. I only need to open the PR on icub_firmware_shared

pattacini:
> Just done. I only need to open the PR on icub_firmware_shared

I meant, to align the required version of icub-firmware-shared here.
I don't see the update.

> I only need to open the PR on icub_firmware_shared

What do you mean? The PR is already there, seemingly.

conf/iCubFindDependencies.cmake
```diff
@@ -65,7 +65,7 @@ checkandset_dependency(OpenCV)
 checkandset_dependency(Qt5)
 
 if(icub_firmware_shared_FOUND AND ICUB_USE_icub_firmware_shared)
-  if(icub_firmware_shared_VERSION VERSION_LESS 1.36.2)
+  if(icub_firmware_shared_VERSION VERSION_LESS 1.36.3)
     message(FATAL_ERROR "An old version of icub-firmware-shared has been detected: at least 1.36.2 is required")
```
pattacini, line 69: Also this line needs to be updated, @MSECode.

pattacini (changes requested):
See inline comments.

pattacini (approved):

valegagge:
We need to keep this issue in draft until the temperature reading feature doesn't pass tests on robots.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -56,6 +57,181 @@ namespace yarp {
 //    PidAlgo_currentInnerLoop =2
 //} PidAlgorithmType_t;
 
+class ITemperatureSensor
```
valegagge, line 60: It is a good idea to implement the "generic interface" `ITemperatureSensor`, but I think you can use the constructor instead of `setTemperatureSensor` function.
In the constructor, you can initialize directly all the specific values of gradient, offset, etc.
In this way, you don't need the parameter `isConfigured`. If the object exists it is also configured.

See the line 1419 of embObjParser.cpp

In addition all the private field could be a const and initialized in the creation pahse.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -56,6 +57,181 @@ namespace yarp {
...
+    double _resolution_pga;
+    double _resolution_tdb;
+
+    bool isConfigured = false;
+
+public:
+
+    virtual void setTemperatureSensor()
+    {
+        _r_1 = 4700;
```
valegagge, line 93: Put the description of the meaning of each parameter.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -56,6 +57,181 @@ namespace yarp {
...
+    virtual double convertTempCelsiusToRaw(const double temperature) override
+    {
+        double res = 0;
+        if (!isConfigured)
+        {
+            yError("Cannot proceed, class paramters not confgured");
+            return res;
+        }
+       	
+        double tmp = (( (_ptc_offset + _ptc_gradient * temperature) / ((double)_r_1 + (_ptc_offset + _ptc_gradient * temperature))) - (double)((double)_r_3 / ((double)_r_2 + _r_3))) * (double)_vcc;
```
valegagge, line 116: From the performance point of view, it is better if this calculation is performed at once in the creation phase. You can save the tmp value in the obj.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -56,6 +57,181 @@ namespace yarp {
...
 //} PidAlgorithmType_t;
 
+class ITemperatureSensor
+{
+
+public:
+    // virtual ~ITemperatureSensor() = default; //TODO: note -> if I use smart pointer I do not need the destructor - theoretically, since it gets deleted when it goes out of scope - 
+
+    virtual void setTemperatureSensor() = 0;
+    virtual double convertTempCelsiusToRaw(const double temperature) = 0;
```
valegagge, line 67: Both `ConvertTempCelsiusToRaw` and `convertRawToTempCelsius` should have the same signature. I suggest returning the converted value.

You could add the sensor type and a method to get it in order to call the conversion functions only on the object different from NONE.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -316,6 +498,13 @@ typedef struct
     float32_t P0;
 } kalmanFilterParams_t;
 
+
+typedef struct
+{
```
valegagge, line 503: Add a comment to explain that these limits are expressed in Raw value

src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
```diff
@@ -973,8 +981,33 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
...
+
+        for (j = 0; j < _njoints; j++)
+        {
+            if (((_foc_based_info[j].temperatureSensorType != motor_temperature_sensor_none ) || (_foc_based_info[j].hasTempSensor != 0)) && ((_temperatureLimits[j].hardwareTemperatureLimit == 0) || (_temperatureLimits[j].warningTemperatureLimit == 0)))
+            {
+                yError() << "In" << getBoardInfo() << "joint" << j << ": inconsistent configuration, please update it. If Temperature limits are not set then TemperatureSensorType must be NONE or not set and/or HasTempSensor must be zero. Aborting...";
+                return false;
+            }
+            
+            //TODO: Add check to _temperatureLimits[j].hardwareTemperatureLimit != 0. If we are here it should be already as this because we are doing check when parsing value from config.
```
valegagge, line 995: Are you sure about this?

src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
```diff
@@ -973,8 +981,33 @@ bool embObjMotionControl::fromConfig_Step2(yarp::os::Searchable &config)
...
+        for (j = 0; j < _njoints; j++)
+        {
+            if (((_foc_based_info[j].temperatureSensorType != motor_temperature_sensor_none ) || (_foc_based_info[j].hasTempSensor != 0)) && ((_temperatureLimits[j].hardwareTemperatureLimit == 0) || (_temperatureLimits[j].warningTemperatureLimit == 0)))
+            {
+                yError() << "In" << getBoardInfo() << "joint" << j << ": inconsistent configuration, please update it. If Temperature limits are not set then TemperatureSensorType must be NONE or not set and/or HasTempSensor must be zero. Aborting...";
+                return false;
+            }
+            
+            //TODO: Add check to _temperatureLimits[j].hardwareTemperatureLimit != 0. If we are here it should be already as this because we are doing check when parsing value from config. 
+            if (_foc_based_info[j].temperatureSensorType == motor_temperature_sensor_pt100)
```
valegagge, line 996: you don't need to check the type of the sensor. Automatically the correct function is called based on the type of the object

src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
```diff
@@ -3269,7 +3321,15 @@ bool embObjMotionControl::getKinematicMJRaw(int j, double &rotres)
     return false;
 }
 
-bool embObjMotionControl::getHasTempSensorsRaw(int j, int& ret)
+bool embObjMotionControl::getTemperatureSensorTypeRaw(int j, int& ret)
```
valegagge, line 3324: Since you added this feature, I suggest returning the type as a string and not as int so it is more clear to the user.

src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
```diff
@@ -4681,14 +4747,28 @@ bool embObjMotionControl::getTemperatureRaw(int m, double* val)
     bool ret = res->getLocalValue(protid, &status);
     if(ret)
     {
-        *val = (double) status.mot_temperature;
+        *val = (double) status.mot_temperature ;
+        if (_foc_based_info[m].temperatureSensorType == motor_temperature_sensor_pt1000)
```
valegagge, line 4751: Also here you don't neet to check the type of the sensor.

valegagge (changes requested):
Hi @MSECode !
Good work! 💪 
I put some suggestions in the review comments.

We can also discuss f2f :)

pattacini (changes requested):
Hi @MSECode 

This PR is going a bit wild.

Can you please edit/squash the commits so to clean the last one containing references to `ergocub` as author?

MSECode:
> n the last one containing references to `ergocub` as author?

Yes, sure. I was already planning to do that, since there's plenty of dirty commits

pattacini:
Hi @MSECode 

Better off squashing again and rebasing on top of the latest `icub-main@devel`, which has received new commits.

There are also conflicts here to address.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -56,6 +57,226 @@ namespace yarp {
...
+
+        _ptc_offset = 100.0;
+        _ptc_gradient = 0.3851;
+        _pga_gain = 2;
+        _vcc = 5;
+        _resolution_pga = 2.048;
+        _resolution_tdb = 32767; 
+
+        _half_bridge_resistor_coeff = (double)_r_3 / (double)(_r_2 + _r_3);
+        isConfigured = true;
```
valegagge, line 116: `isconfigured` is no longer useful. Its value is always true.
Please remove it.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -56,6 +57,226 @@ namespace yarp {
...
+    TemperatureSensorPT100& operator=(const TemperatureSensorPT100& other) = default; // const copy assignment operator
+    TemperatureSensorPT100(TemperatureSensorPT100&& other) = default; // move constructor
+    TemperatureSensorPT100& operator=(TemperatureSensorPT100&& other) = default; // move assignment operator
+
+    ~TemperatureSensorPT100() = default; // destructor
+
+    virtual double convertTempCelsiusToRaw(const double temperature) override
+    {
+        double res = 0;
+        if (!isConfigured)
```
valegagge, line 129: this `if` has no longer sense.

Remove also from PT1000

src/libraries/icubmod/embObjMotionControl/eomcParser.h
```diff
@@ -56,6 +57,226 @@ namespace yarp {
...
+        {
+            yError("Cannot proceed, class paramters not confgured");
+            return res;
+        }
+        
+        double tmp = temperature * ((_resolution_pga) / (_pga_gain * _vcc * (_resolution_tdb + 1)));
+        double den = _ptc_gradient * (_r_2 - _r_2*tmp - _r_3*tmp);
+        res = (tmp * (_r_1*_r_2 + _r_1*_r_3 + _ptc_offset*_r_2 + _ptc_offset*_r_3) / den) + ((_r_3*_r_1 - _r_2*_ptc_offset) / den);
```
valegagge, line 153: Please put `(_r_1*_r_2 + _r_1*_r_3 + _ptc_offset*_r_2 + _ptc_offset*_r_3) / den) + ((_r_3*_r_1 - _r_2*_ptc_offset) / den`in a const field. This small change speeds up a bit the calculation. 

Same for PT1000

valegagge (changes requested):

MSECode:
> Hi @MSECode
> 
> Better off squashing again and rebasing on top of the latest `icub-main@devel`, which has received new commits.
> 
> There are also conflicts here to address.

Rebased and update the icub_firmware_shared version to 1.37.2

src/libraries/icubmod/embObjMotionControl/eomcParser.h
MSECode: Check last commit. Update as requested

src/libraries/icubmod/embObjMotionControl/eomcParser.h
MSECode: Check last commit. Update as requested

src/libraries/icubmod/embObjMotionControl/eomcParser.h
MSECode: Check last commit. Update as requested

pattacini (approved):
Thanks!
Awaiting https://github.com/robotology/icub-firmware-shared/pull/89 to be merged before turning this PR into ready for review.

MSECode:
Hi @valegagge,
check latest commits, if everything is fine you can do final review and then we can merge.
Yesterday I also updated the icub-firmware-build with the latest working versions we tested together.
One doubt is: do I need also to update the foc versions for icub3 and r1 or we can leave them unchanged since they do not have the temperature sensors.

src/libraries/icubmod/embObjMotionControl/eomcParser.h
valegagge: ok

MSECode:
Waiting the other PRs: https://github.com/robotology/icub-firmware/pull/430, https://github.com/robotology/icub-firmware-shared/pull/89, https://github.com/robotology/icub-firmware-build/pull/118 to be merged so that we can move this to ready and let the CR start.

pattacini:
https://github.com/robotology/icub-firmware-shared/pull/89 merged ✅ 
Enabling CI here.

valegagge:
I just merged the https://github.com/robotology/icub-firmware-shared/pull/89.

@pattacini can you put this pr in ready? thanks

pattacini:
> @pattacini can you put this pr in ready? thanks

Already done 😄

PR #918, opened 2023-12-12
embObjAnalogSensor: fix compilation after the removal of EoTheInertials2 service

Nicogene:
This PR fixes a regression of https://github.com/robotology/robots-configuration/issues/595

I think that `embObjAnalogSensor` is kind of deprecated, and for the inertial we use `embObjIMU` also for internal and external mtb3, but since it didn't compile we had to put a patch.
We didn't tested if actually works since it is not used for handling the imu right now

cc @MSECode

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,7 @@ jobs:
               -DENABLE_icubmod_canBusSkin:BOOL=ON \
               -DENABLE_icubmod_canBusVirtualAnalogSensor:BOOL=ON \
               -DENABLE_icubmod_embObjFTsensor:BOOL=ON \
+              -DENABLE_icubmod_embObjAnalogSensor:BOOL=ON \
               -DENABLE_icubmod_embObjIMU:BOOL=ON \
               -DENABLE_icubmod_embObjInertials:BOOL=ON \
               -DENABLE_icubmod_embObjMais:BOOL=ON \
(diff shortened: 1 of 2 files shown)


pattacini (approved):

pattacini:
`embObjAnalogSensor` relies on deprecated I/F.
To be discussed with @marcoaccame and @Nicogene.

/remind December 18

MSECode:
cc: @valegagge

marcoaccame:
> `embObjAnalogSensor`

No, it is not used anymore. Only in some xml files of some old experimental setup. 

See here:

image

pattacini:
Thanks @marcoaccame for reviewing this.

As a result, I think we can:
- Close this PR w/o merging it.
- Remove `embObjAnalogSensor` with a dedicated PR.

cc @Nicogene

Nicogene:
> Thanks @marcoaccame for reviewing this.
> 
> As a result, I think we can:
> 
>     * Close this PR w/o merging it.
> 
>     * Remove `embObjAnalogSensor` with a dedicated PR.
> 
> 
> cc @Nicogene

I am on it!

marcoaccame (approved):
Hi @Nicogene, if the changes you did let the CI work that is fine and the PR can be merged.

Rather, I think we should also remove the device because:

- the `embObjAnalogSensor` is not used anymore by `robots-configuration`
- I am sure that the existing code did nothing before and still does nothing now. 
- It does nothing because the function `bool embObjAnalogSensor::update(eOprotID32_t id32, double timestamp, void* rxdata)` calls `bool embObjAnalogSensor::fillDatOfInertial(void *inertialdata)` that does nothing so it does not update the content of the YARP port.

PR #935, opened 2024-01-22
Bugfix `kbemf` set to 0 by default when not found

sgiraz:
This PR enables the `eomcParser` to parse all parameters within `TRQ_PID_OUTPUT_CURR`, even if `kbemf` is missing.

cc @pattacini @Nicogene @isorrentino @DanielePucci

### DIFF
--- a/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
@@ -903,8 +903,14 @@ bool Parser::parsePidsGroupDeluxe(Bottle& pidsGroup, Pid myPid[])
 
     Bottle xtmp;
 
-    if (!extractGroup(pidsGroup, xtmp, "kbemf", "kbemf parameter", _njoints, false)) return false; 
-    for (int j = 0; j<_njoints; j++) _kbemf[j] = xtmp.get(j + 1).asFloat64();
+    if (!extractGroup(pidsGroup, xtmp, "kbemf", "kbemf parameter", _njoints, false)) 
+    {
+        for (int j = 0; j<_njoints; j++) _kbemf[j] = 0;
+    } 
+    else
+    {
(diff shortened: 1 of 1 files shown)


DanielePucci:
Cc @GiulioRomualdi

src/libraries/icubmod/embObjMotionControl/eomcParser.cpp
```diff
@@ -903,8 +903,14 @@ bool Parser::parsePidsGroupDeluxe(Bottle& pidsGroup, Pid myPid[])
 
     Bottle xtmp;
 
-    if (!extractGroup(pidsGroup, xtmp, "kbemf", "kbemf parameter", _njoints, false)) return false; 
-    for (int j = 0; j<_njoints; j++) _kbemf[j] = xtmp.get(j + 1).asFloat64();
+    if (!extractGroup(pidsGroup, xtmp, "kbemf", "kbemf parameter", _njoints, false)) 
+    {
+        for (int j = 0; j<_njoints; j++) _kbemf[j] = 0;
```
pattacini, line 908: We should print a warning here.

pattacini (changes requested):
Thanks heaps for the fix @sgiraz 
Just one inline comment to address.

pattacini (approved):
Thanks!

randaz81:
Is kbemf not used anymore? Then it's ok to have a default = 0. Otherwise I'd like to say that's a bad software practice to mix mandatory with optional parameters in the same parser block. In the long term things get confused, undocumented and In the past I have seen default value changing without any notice, leading to difficult-to-debug situations. For this reason, at least in yarp, the use default parameters is generally discouraged, and a revision of the configuration files to add the missing parameter is preferred.

pattacini:
Hi @randaz81 

> Is kbemf not used anymore? Then it's ok to have a default = 0.

Indeed, `kbemf` is deprecated and already removed from the config files as per https://github.com/robotology/robots-configuration/pull/599.

Issue #937, opened 2024-01-29
Fix diagnostic messages


valegagge:
### Bug description

In the log reported [here](https://github.com/robotology/icub-tech-support/issues/1726#issue-2105089872), I noticed some weird diagnostic messages that might be worth investigating further. 

1. ` 426.026906 <ERROR> from BOARD 10.0.1.8 (left_leg-eb8-j0_3) time=467s 485m 342u :  MC: generic motor error. (Error is a000000)` .==>  Here the joint ID is missing

3. ` 418.960928 <WARNING> from BOARD 10.0.1.2 (left_arm-eb2-j0_1) time=460s 417m 341u : SYS: a service has detected that some CAN boards are not broacasting anymore. Type of service category is eomn_serv_category_temperatures. Lost can boards on (can1map, can2map) = ([ 13  ], [  ] )`. ==>  Why temperature? it was disabled. And what is the address `13`?

5. `420.013520 <ERROR> from BOARD 10.0.1.2 (left_arm-eb2-j0_1) time=461s 466m 194u : SYS: a service has detected that some CAN boards have stopped transmission. Type of service category is eomn_serv_category_all. Lost CAN boards are on (can1map, can2map) = ([ 2  ], [  ]). Time since last contact: 0 [ms]`. ==>  Why 0 ms? 

### Steps to reproduce

See the ticket https://github.com/robotology/icub-tech-support/issues/1726

pattacini:
A bug related to diagnostic is also:
- #913

MSECode:
Regarding the errors:
- Point 3 is easy to fix --> if seen that in the EMS code we are not passing some of the info we are trying to request on the icub-main part. As shown by the screenshots we are not adding to the last significative bits of the par64 the timing information, as it is done for other errors:

![image](https://github.com/robotology/icub-main/assets/77933235/00e6b355-9510-4918-86c2-300c48a59c8c)

but we are just passing the masks of the CAN boards that are `justLOST` :

![image](https://github.com/robotology/icub-main/assets/77933235/28dcd597-03a7-432c-b5a1-e268da9d4c68)
 

so we just need to add that piece of information to the par64, since `mspassed` can be calculated for all the cases.

MSECode:
Point 2 instead is quite weird, since it is related to this message:
`eocanprotASperiodic_parser_PER_AS_MSG__THERMOMETER_MEASURE` which should not be related to any board, since I suppose there's not any board which currently uses this feature.

[...]

PR #941, opened 2024-02-16
Fix parsing data dimension in lost can boards error

MSECode:
Fix added with this PR:

- Remove parsing time from justLOST canmonitor error, add parsing time to justFOUND and fix canmapping parsing Fix canmapping in can service errors
- Fix datatype and copy-paste typo
- Fix missing joint ID in motor generic error message

### DIFF
--- a/src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
+++ b/src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
@@ -597,7 +597,12 @@ void MotionControlParser::parseInfo()
 
         case eoerror_value_MC_generic_error: //TBD Check print
         {
-            snprintf(str, sizeof(str), " %s (Error is %lx)", m_dnginfo.baseMessage.c_str(), m_dnginfo.param64);
+            uint16_t joint_num = m_dnginfo.param16;
+            m_entityNameProvider.getAxisName(joint_num, m_dnginfo.baseInfo.axisName);
+
+            snprintf(str, sizeof(str), " %s (Joint=%s (NIB=%d) (Error is %lx)", 
+                                        m_dnginfo.baseMessage.c_str(), m_dnginfo.baseInfo.axisName.c_str(), joint_num, m_dnginfo.param64
+                                        );
             m_dnginfo.baseInfo.finalMessage.append(str);
         } break;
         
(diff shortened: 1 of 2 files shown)


valegagge:
Hi @MSECode , 
Good catch! 💪 
We need to be careful about the duplicated code.

I don't know which is the duplicated code `SonarCloud` referred to, but, in any case, is it necessary to avoid the replicated code:

```
            eOmn_serv_category_t serv_category =  (eOmn_serv_category_t)m_dnginfo.param16;
	        uint16_t foundMaskcan1 = (m_dnginfo.param64 & 0x00000000ffff0000) >> 16;
            uint16_t foundMaskcan2 = (m_dnginfo.param64 & 0x000000000000ffff);

            char foundCanBoards1[64] = {0};
            char foundCanBoards2[64] = {0};

            for(int i=1; i<15; i++)
            {
                if ( (foundMaskcan1 & (1<<i)) == (1<<i) )
                {
                    strcat(foundCanBoards1,  std::to_string(i).c_str());
                    strcat(foundCanBoards1, " ");
                }

                if ( (foundMaskcan2 & (1<<i)) == (1<<i) )
                {
                    strcat(foundCanBoards2,  std::to_string(i).c_str());
                    strcat(foundCanBoards2, " ");
                }
            }

            snprintf(str, sizeof(str), "%s Type of service category is %s. CAN boards are on (can1map, can2map) = ([ %s ], [ %s ])",
                m_dnginfo.baseMessage.c_str(),
                eomn_servicecategory2string(serv_category),
                foundCanBoards1,
                foundCanBoards2
            );
            m_dnginfo.baseInfo.finalMessage.append(str);
```

We ought to create a function that parametrizes the string to print in case of `justlost`, `stilllost`, etc..

I guess we can use the same function in the cases ` eoerror_value_SYS_canservices_boards_lostcontact` and `eoerror_value_SYS_canservices_boards_retrievedcontact` because currently they have the same output string and it is necessary to disambiguate the two cases.

valegagge:
Hi @MSECode ,
we could improve a bit more in this way:
```
const int maxstr =64; 
typedef char miastr[maxstr];



void MotionControlParser::miaFunz(uint16_t foundMaskcan, miastr64 foundCanBoards)
{
    for(int i=1; i<15; i++)
    {
        miastr b{};
        if ( (foundMaskcan1 & (1<<i)) == (1<<i) )
        {
            //std::to_chars(aux, aux+2, i);
            snprintf(b, maxstr, "%d ", i);
            strcat(foundCanBoards, b);
        }
    }

}

MotionControlParser::new_private_function (eOmn_serv_category_t serv_category, miastr64 foundCanBoards1, miastr64 foundCanBoards2)
{

            eOmn_serv_category_t serv_category =  (eOmn_serv_category_t)m_dnginfo.param16;
	        uint16_t foundMaskcan1 = (m_dnginfo.param64 & 0x00000000ffff0000) >> 16;
            uint16_t foundMaskcan2 = (m_dnginfo.param64 & 0x000000000000ffff);

            miaFunz(foundMaskcan1, foundCanBoards1);
            miaFunz(foundMaskcan2, foundCanBoards2);
    

}

```

In this way, we reuse the code totally and provide a bit safer code forcing the use of the sting with a fixed size. Moreover, we avoid creating a temporary object on the heap with the to_string function.
What do you think?

valegagge (approved):
well done!
Test already done! see here see https://github.com/robotology/icub-main/issues/937

valegagge:
This PR is in draft, because we are waiting for the https://github.com/robotology/icub-firmware-shared/pull/92.

cc @pattacini

pattacini:
https://github.com/robotology/icub-firmware-shared/pull/92 is merged.
Putting this PR in ready for review to enable CI.

PR #942, opened 2024-02-21
Fix/diagnostic gen error

MSECode:
Updates for generic error translation from bits to string  using this method `motorStatusBitsToString`

correlated to: https://github.com/robotology/icub-main/pull/941 since it's rebased on that and it should be merged after
and it is directly related to: https://github.com/robotology/icub-firmware/pull/472

Related to:
https://github.com/robotology/icub-firmware-shared/pull/94
https://github.com/robotology/icub-firmware/pull/472

### DIFF
--- a/src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
+++ b/src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
@@ -6,10 +6,10 @@
  * BSD-3-Clause license. See the accompanying LICENSE file for details.
  */
 
-#include <string>
 #include "diagnosticLowLevelFormatter_hid.h"
 #include "diagnosticLowLevelFormatter.h"
 #include "EoBoards.h"
+#include "embot_core_binary.h"
 
 
 using namespace Diagnostic::LowLevel;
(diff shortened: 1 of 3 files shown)


pattacini:
Hi @MSECode 

Is this linked to #937?
Please, don't open up PR with a blank body 😉

valegagge:
Hi @MSECode,
I think here you could use the string_view for your map.
Moreover, concatenating the strings in this way means a lot of heap allocation.  
We need to try to reduce them. We can discuss how to do it.
`strcat` is not the best way for safety reasons, but we can find a good solution for it.

PS: I know in the diagnostic parser I used some strings, but I want to improve it to reduce the heap usage anyway.

MSECode:
let's add that point to the f2f discussion as well

conf/iCubFindDependencies.cmake
```diff
@@ -64,7 +64,7 @@ checkandset_dependency(IPOPT)
 checkandset_dependency(OpenCV)
 checkandset_dependency(Qt5)
 
-set(MINIMUM_REQUIRED_icub_firmware_shared_VERSION 1.37.3)
+set(MINIMUM_REQUIRED_icub_firmware_shared_VERSION 1.37.4)
```
pattacini, line 67: Please check this.

pattacini (commented):
Hi @MSECode 

#941 is now merged.
Can you rebase this compare branch?
Especially for what regards the fw-shared dependencies?

To be kept still in draft because of:
- https://github.com/robotology/icub-firmware/pull/472

pattacini:
There are also conflicts on some files.

src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
```diff
@@ -1393,7 +1457,7 @@ void InertialSensorParser::parseInfo()
         case eoerror_value_IS_arrayofinertialdataoverflow:
         {
             
-            uint8_t frame_id = m_dnginfo.param16 & 0x00ff;
+            uint8_t frame_id = m_dnginfo.param16 & 0x0fff;
```
valegagge, line 1460: Good catch!

src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
```diff
@@ -532,6 +532,69 @@ void ConfigParser::parseInfo()
 /******************************************   MotionControlParser   ***************************************************/
 /**************************************************************************************************************************/
 
+// private class functions
+std::string MotionControlParser::motorStatusBitsToString(eOmc_motorFaultState_t motorstatus)
+{
+    // dimesion of the array is hard-code but it must be aligned w/ the union typedef eOmc_motorFaultState_t
+    static const std::array<std::string_view, 32> s_motor_fault_status =
```
valegagge, line 539: use eOmc_motorFaultState_numberof instead of 32

valegagge (approved):
LGTM

MSECode:
I need to update the CMakeLists.txt adding `icub_firmware_shared::embot` to fix compilation

pattacini (approved):

MSECode:
> Awaiting the CI before merging.

Ok,  understood. I leave it running since I did not get any building error locally (since the compiler was probably able to find the embot library as installed on my system ) and I realized the missing of it just from the previous failing of the CI. 
Is there a best practice I can use to simulate the job of the CI and so find any possible building error also locally (I'm thinking if could be useful in this case to use a virgin docker image but let me know what should be the best)?

Another point: regarding the Quality Gate. How should I consider the lines of code interested by a failing in the code analysis that have not been touched by the changes on the PR but still checked by SonarCloud?

pattacini:
>Is there a best practice I can use to simulate the job of the CI and so find any possible building error also locally (I'm thinking if could be useful in this case to use a virgin docker image but let me know what should be the best)?

Docker images are a possibility but somewhat limited to linux envs.
On Windows, there are maybe alternatives to create testing environments locally. One solution could be relying on Windows Sandbox, which in turn can be configured for the problems at hand.

>Another point: regarding the Quality Gate. How should I consider the lines of code interested by a failing in the code analysis that have not been touched by the changes on the PR but still checked by SonarCloud?

We're not handling SonarCloud: no worries

PR #951, opened 2024-03-11
Add MotorTemperaturePublisher tool

MSECode:
Add to `src/tools` a new tool called `motorTemperaturePublisher` which might be helpful for publishing the motor temperatures read from the ETH callback on the software. As things now the module can be started and used on a single robot subpart at the same time. So, if you need to check the temperature of multiple subparts or you start multiple instances of the application or you change the `app/conf/config.ini` file each time defining under `portprefix` the subpart that you need and in `listofjoints` the list of the jonts you desire to read the temperature. 

All other specific info about what the module does and how to build it can be found in the `README.md`


src/tools/motorTemperaturePublisher/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+find_package(YARP REQUIRED)
+# set up our program
+add_executable(${PROJECT_NAME} ./src/MotorTemperaturePublisher.h ./src/MotorTemperaturePublisher.cpp  ./src/main.cpp )
+
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/app/conf/config.ini DESTINATION ${YARP_CONTEXTS_INSTALL_DIR}/${PROJECT_NAME})
+
+# link with YARP libraries
+target_link_libraries(${PROJECT_NAME} ${YARP_LIBRARIES})
+
+install(TARGETS ${PROJECT} DESTINATION bin)
```
pattacini, line 13:
```suggestion
install(TARGETS ${PROJECT_NAME} DESTINATION bin)
```

src/tools/motorTemperaturePublisher/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+cmake_minimum_required(VERSION 3.16)
+project(MotorTemperaturePublisher)
+# find YARP
+find_package(YARP REQUIRED)
```
pattacini, line 4: Better off narrowing down what you need. Something like:
```cmake
find_package(YARP COMPONENTS os sig dev math)
```

src/tools/motorTemperaturePublisher/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+# set up our program
+add_executable(${PROJECT_NAME} ./src/MotorTemperaturePublisher.h ./src/MotorTemperaturePublisher.cpp  ./src/main.cpp )
+
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/app/conf/config.ini DESTINATION ${YARP_CONTEXTS_INSTALL_DIR}/${PROJECT_NAME})
```
pattacini, line 8: This shouldn't be here as app files must go to `icub-main/app`.

src/tools/motorTemperaturePublisher/README.md
pattacini, line 1: We don't use README files.

pattacini (changes requested):
Some points to address:
- Missing copyright in the code files.
- App files need to be moved to `icub-main/app`. Get inspired by how it's done from one example there.
- We don't use README but rather doxygen to publish [documentation](https://robotology.github.io/robotology-documentation/doc/html/).
  - See for example: https://github.com/robotology/icub-main/blob/devel/src/modules/motionCUT/motionCUT.xml.
- We also need to account for the new addition by extending the `src/tools/CMakeLists.txt` file.

src/tools/motorTemperaturePublisher/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 3.16)
+project(MotorTemperaturePublisher)
+# find YARP
+find_package(YARP COMPONENTS os sig dev math)
```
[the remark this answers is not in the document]
MSECode: also `dev` I need

src/tools/motorTemperaturePublisher/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+project(MotorTemperaturePublisher)
```
pattacini, line 1:
```suggestion
# Copyright: (C) 2024 iCub Facility - Istituto Italiano di Tecnologia
# Authors: Jacopo Losi
# CopyPolicy: Released under the terms of the GNU GPL v2.0.

project(MotorTemperaturePublisher)
```

CMake is also code, actually.

pattacini (changes requested):
The very last point is to extend this CMakeLists.txt file:
- https://github.com/robotology/icub-main/blob/master/app/CMakeLists.txt

pattacini (approved):

PR #954, opened 2024-04-04
Fix constructor and destructor of `eth::EthBoards` class

MSECode:
This PR fix the problem seen on windows when launching the `yarprobotinterface` where weird characters were shown on the logs. 
This was due due to a not correct initialization of the variables of the struct `ethboardProperties_t` used by the `LUT` array in `ethBoards.h`. This bad initialization was breaking the copying of new `std::string` vars in the variable `name` of that struct.

Now as shown in the related issue here: https://github.com/robotology/icub-main/issues/913, the prints are fine.

### DIFF
--- a/src/libraries/icubmod/embObjLib/ethBoards.cpp
+++ b/src/libraries/icubmod/embObjLib/ethBoards.cpp
@@ -54,14 +54,31 @@
 
 eth::EthBoards::EthBoards()
 {
-    memset(LUT, 0, sizeof(LUT));
+    // memset(LUT, 0, sizeof(LUT));
+    // Changed memset with for loop initialization due to problem on Windows
+    // which was failing in updating correctly values to LUT vars
+    for (auto& lut : LUT)
+    {
+        lut.clear();
+    }
+    
     sizeofLUT = 0;
 }
 
 
 eth::EthBoards::~EthBoards()
 {
-    memset(LUT, 0, sizeof(LUT));
+    /**
+    * LUT has already been cleaned by the call to
+    * eth::EthBoards::rem(eth::AbstractEthResource* res, iethresType_t type) and
+    * eth::EthBoards::rem(eth::AbstractEthResource* res)
+    * so there's nothing to do here since LUT is already destroyed correctly
+    * but we can keep the clear for syntax reasons
+    */
+    for (auto& l : LUT) 
+    { 
+        l.clear(); 
+    } 
     sizeofLUT = 0;
 }
 
@@ -202,14 +219,9 @@ bool eth::EthBoards::rem(eth::AbstractEthResource* res)
         return false;
     }
 
-    LUT[index].resource = NULL;
-    LUT[index].ipv4 = 0;
-    LUT[index].boardnumber = 0;
-    LUT[index].name = "";
-    LUT[index].numberofinterfaces = 0;
-    for(int i=0; i<iethresType_numberof; i++)
-    {
-        LUT[index].interfaces[i] = NULL;
+    for (auto& l : LUT) 
+    { 
+        l.clear(); 
     }
 
     sizeofLUT--;
--- a/src/libraries/icubmod/embObjLib/diagnosticInfoFormatter.cpp
+++ b/src/libraries/icubmod/embObjLib/diagnosticInfoFormatter.cpp
@@ -6,7 +6,6 @@
  * BSD-3-Clause license. See the accompanying LICENSE file for details.
  */
 
-#include <string>
 #include "diagnosticLowLevelFormatter.h"
 #include "diagnosticLowLevelFormatter_hid.h"
 
--- a/src/libraries/icubmod/embObjLib/diagnosticLowLevelFormatter_hid.h
+++ b/src/libraries/icubmod/embObjLib/diagnosticLowLevelFormatter_hid.h
@@ -11,8 +11,10 @@
 #define __diagnosticLowLevelFormatter_hid_h__
 
 #include <string>
+#include <string_view>
 #include <memory>
 #include <yarp/os/LogStream.h>
+#include <yarp/os/Log.h>
 #include "diagnosticLowLevelFormatter.h"
 #include "EoError.h"
 
--- a/src/libraries/icubmod/embObjLib/ethBoards.h
+++ b/src/libraries/icubmod/embObjLib/ethBoards.h
@@ -81,15 +81,25 @@ namespace eth {
 
         // private types      
 
-        typedef struct
+        typedef struct ethboardProperties_t
         {
-            eOipv4addr_t            ipv4;
-            string                  name;
-            uint8_t                 numberofinterfaces;
-            uint8_t                 boardnumber;
-            AbstractEthResource*    resource;
-            IethResource*           interfaces[iethresType_numberof];
-        } ethboardProperties_t;
+            eOipv4addr_t            ipv4 {0};
+            string                  name {""};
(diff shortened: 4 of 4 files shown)


src/libraries/icubmod/embObjLib/ethBoards.cpp
```diff
@@ -54,15 +56,31 @@
...
 eth::EthBoards::~EthBoards()
 {
-    memset(LUT, 0, sizeof(LUT));
-    sizeofLUT = 0;
+    /**
+    * LUT has already been cleaned by the call to
+    * eth::EthBoards::rem(eth::AbstractEthResource* res, iethresType_t type) and
+    * eth::EthBoards::rem(eth::AbstractEthResource* res)
+    * so there's nothing to do here
+    */
```
pattacini, line 83: Are the method `rem` called also while deleting an object of this class?

src/libraries/icubmod/embObjLib/ethBoards.cpp
```diff
@@ -25,6 +25,8 @@
 // --------------------------------------------------------------------------------------------------------------------
 
 #include "ethBoards.h"
+#include <yarp/os/LogStream.h>
+#include <yarp/os/Log.h>
```
pattacini, line 29: Why was this necessary?

pattacini (changes requested):
Thanks @MSECode for the fix 👍🏻 

I'm not the author of the touche modules so I'll leave the review to them. Just dropped some inline non-blocking comments.

I've got only one main point to address:
- Since this is a fix, it should land on `master` rather than `devel`. 

Could you rebase the PR accordingly?
Changing the base branch of the PR is not enough, unfortunately.

src/libraries/icubmod/embObjLib/ethBoards.cpp
```diff
@@ -54,15 +56,31 @@
 
 eth::EthBoards::EthBoards()
 {
-    memset(LUT, 0, sizeof(LUT));
+    // memset(LUT, 0, sizeof(LUT));
```
marcoaccame, line 59: the `memset()` was a quick way to do initialization. If some compiler complains you can drop any code in the constructor and place initialization in the `ethBoards.h` file by improving the declaration of `class EthBoards` as follows:

```C++

        use member inialization

        struct ethboardProperties_t
        {
            eOipv4addr_t            ipv4 {0};
            std::string             name {""};
            uint8_t                 numberofinterfaces {0};
            uint8_t                 boardnumber {0};
            AbstractEthResource*    resource {nullptr};
            IethResource*           interfaces[iethresType_numberof] {nullptr};
            void clear() 
            { 
                // do what you need 
            }
        };

 and also ...

        size_t sizeofLUT {0};
        ethboardProperties_t LUT[EthBoards::maxEthBoards] {};

```

with the above code you don't need anything in the ctor
```C++
eth::EthBoards::EthBoards()
{
    // nothing is required because LUT and sizeofLUT are already OK. 
    // BUT i just do the following anyway:
   for (auto& l : LUT)
   { 
       l.clear();
   }
    sizeofLUT = 0;
}
``` 

One more thing: 

you may also change `bool eth::EthBoards::rem(eth::AbstractEthResource* res)` to call
`LUT[index].clear();` instead of these [lines](https://github.com/robotology/icub-main/blob/devel/src/libraries/icubmod/embObjLib/ethBoards.cpp#L205-L213).

src/libraries/icubmod/embObjLib/ethBoards.cpp
marcoaccame: OK, what happens if we call the dtor w/out calling every `rem()` ? 
Probably nothing because LUT is destroyed by its default dtor that works OK. 

In any case, if you want you can keep the clear()

```C++
eth::EthBoards::~EthBoards()
{
    // nothing is required because LUT is destroyed correctly
    // BUT i just do the following anyway:
   for (auto& l : LUT)
   { 
       l.clear();
   }
    sizeofLUT = 0;
}
```

marcoaccame (changes requested):

src/libraries/icubmod/embObjLib/ethBoards.cpp
MSECode: Yeh, right, I've added those 2 lines in ethBoards.cpp for debug reason. I'll remove them from this PR.

src/libraries/icubmod/embObjLib/ethBoards.cpp
MSECode: Exactly, as you say @marcoaccame nothing happens, since LUT has been already destroyed by the default dtor when calling `rem()`

src/libraries/icubmod/embObjLib/ethBoards.cpp
MSECode: Ok, I'll check this, update and see if it is still fine the execution of the code in Win

pattacini (approved):
Fine by me 👍🏻 
As soon as you put the PR in ready for review and the CI passes, I'll merge it, @MSECode

MSECode:
> Fine by me 👍🏻 As soon as you put the PR in ready for review and the CI passes, I'll merge it, @MSECode

I checked again if YRI is running happily and it is, so if for @marcoaccame the changes are ok too, the PR can be moved in Ready for Review.

marcoaccame:
> > Fine by me 👍🏻 As soon as you put the PR in ready for review and the CI passes, I'll merge it, @MSECode
> 
> I checked again if YRI is running happily and it is, so if for @marcoaccame the changes are ok too, the PR can be moved in Ready for Review.

it is ok by me, thanks

pattacini:
Put in ready for review 👍🏻

pattacini:
Hi @MSECode 

CI is failing.

pattacini (changes requested):
Hi @MSECode 

CI is failing.

pattacini:
The error caught is:
```console
D:\a\icub-main\icub-main\src\libraries\icubmod\embObjLib\ethBoards.h(85,9): error C7626: unnamed class used in typedef name cannot declare members other than non-static data members, member enumerations, or member classes [D:\a\icub-main\icub-main\build\src\unittest\unittest.vcxproj]
```

According to https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5208?view=msvc-170#example, a possible workaround could be giving a fictitious name to the typedef struct.

```c++
typedef struct ETHprop_NamedType
        {
            eOipv4addr_t            ipv4 {0};
            string                  name {""};
            uint8_t                 numberofinterfaces {0};
            uint8_t                 boardnumber {0};
            AbstractEthResource*    resource {nullptr};
            IethResource*           interfaces[iethresType_numberof] {nullptr};

            void clear()
            {
                ipv4 = 0;
                name = "";
                numberofinterfaces = 0;
                boardnumber = 0;
                resource = nullptr;
                for (auto& itf : interfaces) {itf = nullptr;}
            };
        } ethboardProperties_t;
```

valegagge (approved):
LGTM

marcoaccame:
> The error caught is:
> 
> ```
> D:\a\icub-main\icub-main\src\libraries\icubmod\embObjLib\ethBoards.h(85,9): error C7626: unnamed class used in typedef name cannot declare members other than non-static data members, member enumerations, or member classes [D:\a\icub-main\icub-main\build\src\unittest\unittest.vcxproj]


something like the following without the `typedef` will compile:
```C++
struct ethboardProperties_t { .... };
```

pattacini (approved):

Issue #959, opened 2024-05-07
Fix the motor temperature spike filtering  in embObjMotionControl


valegagge:
### Bug description

In embObjMotionControl, a temperature monitor has been implemented to provide a warning to the user when the temperature goes beyond the warning threshold defined in the configuration file.
Since the motor temperature reading suffers from disturbance there are some spikes that have not be take in consideration in the check of the warning threshold.

In addition, the current fw sends some negative values to encode some error cases. Also, this should be removed from the warning check.

### Steps to reproduce

Just read the motor temperature. (In some motors the readings are noisier than others)

### Expected behavior

In the log, there are any warnings related to the spikes of temperature

valegagge:
Just a tip: for debugging and making the work, you can create an ad hoc fw that sometimes sends a spike.
Of course, the final test has to be done on the robot

MSECode:
After developing a first version of the filtering in the sw I've tested it by generating synthetic data with the aim of simulating a continuous rising of the temperatures with a pyramidal pattern. Basically, starting from 0 degree I've then risen the temperature of about 10 degree, so that the threshold between temperature at time interval `i` and `i-1` (20 degree) it is not overcome. Then, after some steps I've instead increment the temperature to a higher value (around 110/120 degree in order to simulate a spike). And finally I've decreased again the temperature value.
Part of the described behavior is shown in this image

![image](https://github.com/robotology/icub-main/assets/77933235/ae62821e-714c-4f78-a615-635960a90c47)

Doing so, and looking at the logs of the YRI, it is possible to see that we do not have any warning related to the generated spike but the warning are just related to the overcoming of the warning temperature limit (set now to 40 degree just for test purposes) when the delta between current and previous temperature is lower than a defined threshold.

![image](https://github.com/robotology/icub-main/assets/77933235/a752908f-635b-4f3f-adf8-64738bc549a1)


cc: @valegagge

MSECode:
/reminder May, 21, 2024

@valegagge @MSECode

MSECode:
## WIP
 my branch 

![Screenshot from 2024-05-21 19-33-08](https://github.com/robotology/icub-main/assets/77933235/dae5d901-276a-4eb0-9fcd-9e451f6252cc)

MSECode:
PR opened here:

https://github.com/robotology/icub-main/pull/961

pattacini:
PR merged ✅ 
Probably, this issue can be closed as complete.

cc @MSECode @valegagge

PR #961, opened 2024-05-22
Update filtering of spikes in temp reading

MSECode:
Completing the work done in https://github.com/robotology/icub-main/issues/959, the spikes in the temperature readings are 
now filtered by software.
This means that if the delta between current and previous temperature is higher than a threshold (which is currently defined in sw and by default is 20 Celsius degree) that new temperature read is considered as a spike and therefore, even if its value is higher than the `warning temperature threshold` or the `error temperature threshold`, it is discarded and not considered and the previous temperature is not updated.
Moreover, we have added to the `Watchdog` class a method for setting the threshold, regarding the maximum number of erroneous values read before sending error or warning, depending on the transmission rate of the ethernet packets.
Therefore, depending on the transmission rate defined in the configuration of the robot, we are setting the watchdogs for overcoming the warning threshold on the temperature reading and for overcoming the limit on the negative temperature values (which means that we got an error in the reading) so that it is always constant and independent to the transmission rate of the ethernet packets.
By default, it is set so that we rise the warning or error when we are getting erroneous temperature values (which are not spikes) for at least 60 seconds.

### DIFF
--- a/src/libraries/icubmod/embObjLib/abstractEthResource.h
+++ b/src/libraries/icubmod/embObjLib/abstractEthResource.h
@@ -53,6 +53,7 @@ namespace eth {
             string              boardtypeString;
             string              boardnameString;
             string              firmwareString;
+            uint8_t             txROPratedivider;
         };
 
         //AbstractEthResource();
--- a/src/libraries/icubmod/embObjLib/diagnosticInfo.cpp
+++ b/src/libraries/icubmod/embObjLib/diagnosticInfo.cpp
@@ -34,6 +34,9 @@ void TimeOfInfo::toString(std::string &str_toi)
 
 void EmbeddedInfo::printMessage()
 {
+    if (finalMessage.size() == 0)
+        return;
+   
    std::string str_toi;
    timeOfInfo.toString(str_toi);
 
--- a/src/libraries/icubmod/embObjLib/ethResource.cpp
+++ b/src/libraries/icubmod/embObjLib/ethResource.cpp
@@ -129,6 +129,7 @@ bool EthResource::open2(eOipv4addr_t remIP, yarp::os::Searchable &cfgtotal)
     properties.ipv4addressingString = brddata.properties.ipv4addressingstring;
     properties.boardtypeString = brddata.properties.typestring;
     properties.boardnameString = brddata.settings.name;
+    properties.txROPratedivider = txconfig.txratedivider;
 
 
     eth::EthMonitorPresence::Config mpConfig;
--- a/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
+++ b/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp
@@ -8,12 +8,14 @@
 */
 
 //#include <yarp/dev/CanBusInterface.h>
-#include <yarp/os/Bottle.h>
-#include <yarp/os/Time.h>
+// system std include
 #include <string.h>
 #include <iostream>
 #include <cmath>
 
+// yarp include
+#include <yarp/os/Bottle.h>
+#include <yarp/os/Time.h>
 #include "embObjMotionControl.h"
 #include <ethManager.h>
 #include <FeatureInterface.h>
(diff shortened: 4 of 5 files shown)


pattacini (approved):

S-Dafarra:
> By default, it is set so that we rise the warning or error when we are getting erroneous temperature values (which are not spikes) for at least 60 seconds.

Hi there, just as a curiosity. Is there a way to get these warnings from code?

MSECode:
I'm not sure if I'm replying to what u need. Anyway, warnings are just related to reading a temperature higher than the warning threshold when that increment it is not a spike (where in this case a spike is defined when the delta between current and previous temperature is higher than 20 Celsius degree). Moreover, the first time temperature goes over the warning threshold a warning message is sent. Then, in order to not flooding the YRI, we send the next warning message if the temperature stays continuously above the threshold for at least 60 seconds. Timing definitions have been set here for now: https://github.com/robotology/icub-main/blob/5410ab082d5cf95109354e6cfd809a3324bc1d82/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.h#L111-L182
Instead, when you read negative readings, you check meaning of Celsius and Raw value in the error message from this table
https://icub-tech-iit.github.io/documentation/temperature_sensors/software/dataflow/#error-handling

S-Dafarra:
I meant that if after 60s the joint goes in HF, I would love to have some warning before, that can be caught not just by reading at the YRI output (which is virtually impossible during experiments), but maybe via some API calls so that we can inform the user/controller accordingly.

MSECode:
The HF is entirely managed by the 2FOC. This part in icub-main has been done to filter all the spikes that can happen due to noisy readings. Those parameters can be eventually adjusted depending on the need. 
Currently there are the APIs for retrieving the instantaneous value of temperature. We can think to even something else if necessary. Or it may be possible to update the module `motorTemperaturePublisher` to let it send advise to the user.
The idea of informing the low level controller can be problematic. We can think to inform the user, but in this case where should we send this "warning". What do u actually need for working fine? What should these APIs make to solve the problem? 
@valegagge, do you have some other ideas?

S-Dafarra:
> The idea of informing the low level controller can be problematic

Just to clarify, I was referring more to a "high-level" controller, like the one controlling the walking.




> Or it may be possible to update the module `motorTemperaturePublisher` to let it send advise to the user.

I don't know about this since the ``motorTemperaturePublisher`` is not something that runs with the robot.



> We can think to inform the user, but in this case where should we send this "warning". What do u actually need for working fine? What should these APIs make to solve the problem?

Maybe a dedicated interface with a defined list of warning codes? I am pretty open to discussion here. Here some possible things I am imagining:
- the ``yarpmotorgui`` starts blinking on the joint that is about to fault
- a module starts making some noise, maybe also sending notifications to the OS (it is possible with QT, see https://doc.qt.io/qt-5/qtwidgets-desktop-systray-example.html)
- the walking controller goes in a fault safe configuration (maybe stopping the walking and going in a joint configuration that is safe in case of sudden HF)

MSECode:
Oks, that's clear. Those points seem good. I'll discuss with the team about an implementation such that.

valegagge:
Hi @S-Dafarra,
as we already discussed some time ago, the high-lever controller, such as the walking controller, should read the current motor temperature and compare it to the warning threshold. (Both information is available on IMotor interface). If the temperature exceeds the threshold should put the robot in a safe position or reduce the load on the motor with high temperature.

If I understand correctly, you are asking to signal in some way the warning state also on the `yarpmotorgui`.  

The work done in this PR had the purpose of avoiding `false-positive warning` to get a cleaner log.

S-Dafarra:
> If I understand correctly, you are asking to signal in some way the warning state also on the `yarpmotorgui`.

I was referring to the case in which there are measurement errors for more than 60s. I was just wondering in which case, the user code can be informed about the potentially imminent fault.

valegagge:
> > If I understand correctly, you are asking to signal in some way the warning state also on the `yarpmotorgui`.
> 
> I was referring to the case in which there are measurement errors for more than 60s. I was just wondering in which case, the user code can be informed about the potentially imminent fault.

The fault on error readings happens after 10 seconds. We had planned to fix the error reading issue by hardware. I'll update you ASAP. Stay tuned

Issue #962, opened 2024-05-29
Crash of `embObjMotionControl` on closure of `yarprobotinterface`


Nicogene:
### Bug description

Today w/ @martinaxgloria we noticed on `iCubGenova11` that w/ `icub-main` on latest devel `yarprobotinterface` crashes on closure.
The faulty plugin is `embObjMotionControl`.

### Steps to reproduce

Run `yarprobotinterface` and close it.

### Expected behavior

It does not crashes in closure

### Additional context

This issue is probably breaking the `robots-configuration` CI
- https://github.com/robotology/robots-configuration/actions/runs/9284133798/job/25545849005

cc @valegagge @MSECode @pattacini @marcoaccame

valegagge:
Hi @Nicogene and @martinaxgloria , 
 Are you on top of devel? are you running YRI with the configuration on devel?

thanks heaps

Nicogene:
Yes both on latest devel, also `icub-firmware-shared`

MSECode:
I noticed it as well. One thing that I saw by working on different setups is that:
- on `ergoCub001`, which is slightly behind on the latest updates with respect to `devel` (on multiple repositories part of the superbuild), I did not notice any  `segmentation fault error` in closure
- on other setups, such as our upper-arm, and on `ergoCub002`, which are instead aligned with the latest updates, I noticed the `segmentation fault error` multiple times while closing the YRI.

It would be possible to start to analyze at which point in the git history we started to have such error.

Nicogene:
Looking at the [history](https://github.com/robotology/icub-main/commits/master/src/libraries/icubmod/embObjMotionControl), there are two potential commits:
- 03/2024 -> https://github.com/robotology/icub-main/commit/6376390d867d57ba4ae4e55fab4467a9a256b1e0
- 05/2024 -> https://github.com/robotology/icub-main/commit/5410ab082d5cf95109354e6cfd809a3324bc1d82

pattacini:
The crash was caught also by the latest robots-configuration CI:
- https://github.com/robotology/robots-configuration/actions/runs/9284133798/job/25545849005

valegagge:
> Looking at the [history](https://github.com/robotology/icub-main/commits/master/src/libraries/icubmod/embObjMotionControl), there are two potential commits:

The history you posted is for master and to for devel 😸 
> 
[...]

PR #963, opened 2024-05-30
Fix bug when removing ethResource from LUT in clear method

MSECode:
This PR fix the bug described here: https://github.com/robotology/icub-main/issues/962 that caused a core dumping when closing the resources for a specific device (when multiple devices are enabled)
 Basically, at the removal of the first resource we were basically going to remove all the open resources.


Nicogene (approved):
Great!

pattacini (approved):
Great! Thanks.

PR #968, opened 2024-06-10
Modify the parsing of the aksim2 diagnostic error

MSECode:
This PR modify slightly the parsing of the aksim2 diagnostic errors adding the number of jomo (joint-or-motor) in board and the number of errors of each type we got in the predefined time.

This is linked to the modifications integrated in https://github.com/robotology/icub-firmware/pull/500

### DIFF
--- a/src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
+++ b/src/libraries/icubmod/embObjLib/diagnosticInfoParsers.cpp
@@ -11,6 +11,8 @@
 #include "EoBoards.h"
 #include "embot_core_binary.h"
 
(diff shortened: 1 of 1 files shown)


valegagge (changes requested):
Hi @MSECode,

the code is ok, but for love of good quality code I suggest collapsing the 4 hw-error cases in only one: the only difference among them is the name of the error, but is already reported in the default string. Instead, you can use the generic word `error`. In this way, we remove the duplicated code.

MSECode:
Update done. Asked re-checking

valegagge (approved):
OK! thanks!

pattacini (approved):

PR #1003, opened 2025-01-08
Remove debug lines for calibration14 and fix temperatures timestamp

MSECode:
This PR fixes some lines of code:
- First of all we removed the debug lines for calibration14 in embObjMotionControl::setCalibrationParametersRaw()
- Moreover we fix the timestamps in log for temperature errors using relative time from construction of embObjMC device 
- Finally we change yError to yWarning when failing to read temperatures considering that in that case the system won't stop

cc: @valegagge


valegagge (approved):
LGTM

pattacini (approved):

PR #1006, opened 2025-01-13
Remove workaround for paramparserGenerator with rawvalues

MSECode:
Removed the workaround introduced for allowing the correct building of the files related to the rawValuesPublisher when using a YARP version older than 3.10.
Now that YARP 3.10.x has been officially released we can enable the automatic generation of the files removing the workaround.


src/libraries/icubmod/fakeRawValuesPublisher/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ yarp_prepare_plugin(fakeRawValuesPublisher
                     GENERATE_PARSER)
 
 if(ENABLE_fakeRawValuesPublisher)
-  option(ALLOW_DEVICE_PARAM_PARSER_GENERATION "Generate the param parser for fakeRawValuesPublisher device" OFF)
+  option(ALLOW_DEVICE_PARAM_PARSER_GENERATION "Generate the param parser for fakeRawValuesPublisher device" ON)
```
pattacini, line 15: Keep it OFF.

pattacini (changes requested):

pattacini (approved):

PR #1023, opened 2025-06-04
Add further check for eomn_serv_MC_mc4plusmais device

martinaxgloria:
As per title, this further check will allow the group [`MAIS`](https://github.com/robotology/robots-configuration/blob/1d292659dc4777b5d5c20d2a702fe8968ea9f8ef/iCubGenova11/hardware/motorControl/right_arm-eb28-j8_11-mc_service.xml#L29-L31) within the motorControl service XML file to be parsed only when the device is of type `eomn_serv_MC_mc4plusmais`, otherwise will be ignored by the parser.


pattacini (approved):
Waiting the CI before merging.

PR #1043, opened 2025-09-11
[iKin] - Introduce iCub v2.10

martinaxgloria:
As per title


martinaxgloria:
Once I had the urdf of `iCubGazeboV2_10`, I tested this PR in Gazebo using iKinGazeCtrl. In particular, I launched the module `iKinGazeCtrl --context gazeboCartesianControl --from iKinGazeCtrl.ini --head_version v2.10` with and without IMU. Then, I attached to the rpc port and I gave the command to look to some (x,y,z) points in the 3D available space. 

- `without` IMU

   https://github.com/user-attachments/assets/d58c198c-873f-4187-8307-f3732cb28bd1


- `with` IMU 

   https://github.com/user-attachments/assets/2e211392-391c-49b0-8634-64550e26635b


As suggested by @pattacini, I also tried to bind the neck joints and increase the velocity of the eyes as you can see in the video below:


https://github.com/user-attachments/assets/0339a08b-efc9-42e7-808b-e8981b46de84

pattacini:
Need to verify what happens to v3.0 though.

pattacini (changes requested):
Version 3.0 still needs to be correctly handled at the body level.
At first sight, it doesn't seem that it's managed for the eyes.

pattacini (approved):
Thanks heaps for the fix!

PR #1065, opened 2026-06-05
Fix minor error for parsing axis names for AMO diagnostic parser

MSECode:
This PR fixes 2 bugs we found while analyzing the diagnostic messages:
- first: we missed to parse the axis name from the base info before adding it to the diagnostic string
- second: we introduced a wrong bite shifting producing an error in the visualization of the right error code for the specific encoder


pattacini (changes requested):
We shall target the master branch.

pattacini (approved):
Thanks!

PR #1067, opened 2026-08-05
Fixed CMakeLists files that were pointing to the wrong output folder

elandini84:
They were pointing towards build/share/yarp which makes it difficult (or better, less intuitive) to use the plugins without using make install


pattacini (approved):
CI failure is unrelated.

PR #1074, opened 2026-09-01
Yarp master compatibility fixes

elandini84:
This quite large PR fixes a set of issues that some classes have with Yarp master branch. These are the classes:

- `embObjFTsensor`
- `embObjIMU`
- `embObjMais`
- `embObjSkin`
- `perceptiveModels/sensors`
- `perceptiveModels/springyFingers`
- `SkinManager/compensator`
- `SkinMeshClient`

This PR is flagged as **DRAFT** because is pretty big. If there are any issue with it, It can be easily split in smaller ones

### DIFF
--- a/src/libraries/icubmod/embObjFTsensor/embObjFTsensor.cpp
+++ b/src/libraries/icubmod/embObjFTsensor/embObjFTsensor.cpp
@@ -330,9 +330,10 @@ eth::iethresType_t embObjFTsensor::type()
 }
 
 // ---------------------- ITemperatureSensors --------------------------------------------------------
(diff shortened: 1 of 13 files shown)


valegagge:
Thanks for your contibution @elandini84 !!!
Did you already test these changes, even if they seem regarding changing of interfaces?

elandini84:
> Thanks for your contibution @elandini84 !!!
> Did you already test these changes, even if they seem regarding changing of interfaces? 

I just verified that they would build correctly. But talking with @randaz81 I found out that the changes in the interfaces I tackled will be released with yarp 4.1. So it is probably better to leave them out for the moment.

